PR #7, opened 2026-06-16
builtin sweep: fidelity fixes + rg removal (model-consensus triage)

tobert:
After doing over sed and awk, I asked an Opus to sweep over all the rest of the builtins looking for footguns and mismatches from "muscle memory" of models. It did this by calling Claude, Deepseek, and Gemini model variants and asking them to produce commands. Then those were compiled into a list and tested with kaish. This PR is landing all the critical fixes that came out of that work.

In addition, this drops the `rg` builtin. It isn't really ripgrep when it's in kaish, and I want to focus on making regular grep work really well before considering it again. There is also a note in docs/issues.md to consider adding some rg features to kaish-glob so it can do file type matching and some other conveniences, then offer them throughout kaish as usual.

-@tobert

--

Breadth-first, model-consensus triage of the non-awk core-util builtins, then a
burn-down of the divergences it surfaced. Reference of correctness = cross-model
banked consensus, **not** GNU coreutils. Every CODE fix lands with a kernel-routed
test (`crates/kaish-kernel/tests/builtin_fidelity_tests.rs`); docs + CHANGELOG ride
each commit; each chunk was DeepSeek-reviewed (kaibo).

## Landed

**Stdin (P1.0 — was a release blocker)**
- Lazy, pipe-backed stdin for `kaish -c`/script: a command that doesn't read stdin
  (`echo`) returns immediately even on an open never-EOF pipe (`sleep 10 | kaish -c
  'echo hi'` no longer hangs). New `Kernel::execute_with_pipe_stdin(_streaming)` seam
  alongside the eager `ExecuteOptions::with_stdin(String)`.
- Fixed every stdin consumer that read the raw `ctx.stdin` field and so missed the
  lazy pipe: external commands in non-first pipeline stages, scatter/gather, and
  `patch`/`write`/`kaish-validate`.

**Fidelity fixes (P2/P3) — all were silent-wrong or hard-error before**
- `tail -n +N` (from line N) · `head -n -N` (all but last N)
- `cut -f` passthrough for non-delimited lines + `-s` · `split --limit` field cap
- `jq -c` compact · `wc` format + newline-counting (`wc -l` counts newlines)
- `tr -c`/`--complement` · `sort -V` version sort

**Format / message / docs (P4)**
- Trailing-newline policy applied to tac/base64/xxd (decode/reverse stay raw)
- `patch` follows `patch(1)` (no `N changes applied` line)
- Comma-bearing argv (`cut -f1,3`) gets a quote-teaching message, not "token pasting"
- `diff` unified-default documented

**BREAKING**
- The `rg` builtin is **removed** (kaish's 80%-rule: one `grep`). pcre2 feature +
  grep-pcre2 dep dropped. Accepted-lost rg-only flags listed in CHANGELOG; a system
  `rg` on PATH still runs externally. Follow-up (issues.md): port `--type`/`--hidden`
  into kaish-glob.

## Deferred
- `tee`/`patch` latch+trash gating → `docs/issues.md` (write-model design; pairs with
  the deferred `sed -i`).

Sweep scaffolding (`docs/builtin-sweep/`) deleted at sign-off; durable record is the
code, the fidelity tests, the CHANGELOG, and the help.


tobert:
## Code-review pass (high-effort, multi-agent) — findings & resolution

After the branch was marked ready, a high-effort review ran 8 independent finder
angles (3 correctness · 3 cleanup · altitude · conventions) over `main...HEAD`,
each verified. It caught a **critical data-loss bug** that the per-chunk reviews
during development missed — because it lives in the *frontend* (`kaish-repl`),
below the kernel API the unit tests exercise. Recording here for posterity.

### Fixed (commit `9e3c330`)

| Sev | Finding | Fix |
|-----|---------|-----|
| 🔴 **CRITICAL** | The lazy-stdin bridge thread used `PipeWriter::write_bytes` — a single `poll_write` that writes at most one ~64 KiB pipe-buffer fill and returns a **short count** — and ignored the count. Any piped stdin >64 KiB was **silently truncated** (`head -c 200000 … \| kaish -c cat` → 118080 bytes). | Use `write_all` (loops until every byte lands). |
| 🟠 **HIGH** | The non-interactive output callback printed `text_out()`, which lossy-decodes a `Bytes` result to U+FFFD — binary output corrupted (`printf '\xff\xfe\x00' \| kaish -c cat` → 7 bytes, not 3). | Write `out_bytes()` raw for binary results. |
| 🟡 **MED** | In `try_execute_external`, the detached pipe→child-stdin task was aborted only on the capture path, not the two interactive `inherit_output` returns → task + upstream-pipe leak. | Single abort-on-drop guard covering every exit path. |

Regression tests spawn the **real binary** (the layer the bugs hid in):
`crates/kaish-repl/tests/cli_stdin_tests.rs` — 200 KiB no-truncation through `cat`
and a pipeline, binary round-trip through `cat` and `base64 -d`.

### Deferred to `docs/issues.md` (real but niche / pre-existing)

- Piped stdin isn't shared across `kaish -c 'echo x; cat'` statements — a
  non-reading first statement drops the reader, so a later `cat` sees nothing. A
  shell shares fd 0; this needs a stdin-lifecycle change. Pre-existing within the
  branch.
- `head -n -0` → empty instead of whole file — `-0` loses its sign at the lexer.
  Waived (genuinely obscure, not fixable without lexer work).

### Cleanup / altitude findings — assessed, no action

- The dual `String` (`ExecuteOptions::with_stdin`) vs `PipeReader`
  (`execute_with_pipe_stdin`) stdin seam is **intentional** — a ready buffer for
  embedders vs a lazy byte-clean stream for the CLI.
- The "duplicate" stdin-copy task in `dispatch.rs` is `#[cfg(test)]`-only
  (`BackendDispatcher`); the production path is `kernel.rs`.

**Takeaway:** the per-chunk DeepSeek reviews validated each fix in isolation but
couldn't see a frontend/process-level bug. Process-level regression tests now
guard the `write_bytes`-vs-`write_all` and `text_out`-vs-`out_bytes` traps.

`cargo test --all` + `cargo clippy --all --all-targets` green.

PR #14, opened 2026-06-18
test(grep): pin -n with -A/-B context line numbering + separators

tobert:
## What

Adds `test_grep_line_numbers_with_context`, exercising `-n` together with `-A`/`-B` context — a combination no existing test covered (we had `-n` alone and context alone).

## Why

`-n`'s interaction with context is where GNU compatibility lives: matched lines use the `:` separator, context lines use `-`, and every emitted line must be numbered. Nothing pinned that, so a regression in either the numbering or the separator logic would have passed silently.

## Test

Matches `two` in `/lines.txt` (`line one\nline two\nline three\nfour`) with `-n -B1 -A1` and asserts exact output:

```
1-line one
2:line two
3-line three
```

Exact-equality so it fails loudly on any drift.

- `cargo test -p kaish-kernel --lib grep` — 44 passed
- `cargo clippy --all-targets` — clean

### DIFF
--- a/crates/kaish-kernel/src/tools/builtin/grep.rs
+++ b/crates/kaish-kernel/src/tools/builtin/grep.rs
@@ -1564,6 +1564,28 @@ mod tests {
         assert!(result.text_out().contains("line three"));
     }
 
+    /// `-n` with `-B`/`-A` context must number every emitted line and keep the
+    /// GNU separator convention: matched lines use `:`, context lines use `-`.
+    #[tokio::test]
+    async fn test_grep_line_numbers_with_context() {
+        let mut ctx = make_ctx().await;
+        let mut args = ToolArgs::new();
+        args.positional.push(Value::String("two".into()));
+        args.positional.push(Value::String("/lines.txt".into()));
+        args.flags.insert("n".to_string());
+        args.named.insert("before-context".to_string(), Value::Int(1));
+        args.named.insert("after-context".to_string(), Value::Int(1));
+
+        let result = Grep.execute(args, &mut ctx).await;
+        assert!(result.ok());
+        // /lines.txt = "line one\nline two\nline three\nfour"; match is line 2.
+        // Before/after context get `-`; the match gets `:`.
+        assert_eq!(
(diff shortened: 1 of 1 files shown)


tobert:
## kaibo review (cast: deepseek — explorer `deepseek-v4-flash`, synth `deepseek-v4-pro`)

**Verdict: correct and ready to land.** kaibo traced the assertion against the real rendering code rather than taking it on faith.

| Question | Verdict |
|---|---|
| Expected output correct? | ✅ Exact match with `render_events` |
| Exercises genuine rendering path? | ✅ Hits `grep_lines_structured` → `render_events`; context flags force the whole-buffer path, skipping the streaming shortcut |
| Fixture + match line correct? | ✅ `"two"` matches only line 2; lines 1 and 3 are the correct before/after context |
| Gaps? | `-C` not a gap (resolves to the same `GrepOptions` + code path). `--` break + `-n` and multi-file + context + `-n` are possible future tests, out of scope here. |

### Detail
- **Separators:** the `prefix(line_num, sep)` closure (`grep.rs:1013–1025`) emits `':'` for `SearchEvent::Match` (`grep.rs:1055`) and `'-'` for all `ContextKind` variants (`grep.rs:1076–1078`) — so `1-line one`, `2:line two`, `3-line three` is exactly right.
- **Path fidelity:** the `before-context`/`after-context` named keys are the same canonical keys the kernel dispatcher binds; `to_argv()` round-trips them to `--before-context=1`/`--after-context=1`, clap parses, `parse_context()` resolves to `Some(1)`. Because context is `Some`, the streaming shortcut (`grep.rs:358–359`) is skipped, so the test genuinely hits the `render_events` separator logic. Consistent with the existing direct-`execute()` convention used by all grep tests.

### Suggested follow-ups (not blocking, deferred)
- multi-file + context + `-n` (filename prefix interacting with line-number prefix: `file.txt:2:line two` for matches, `file.txt-1-line one` for context)
- `--` context-break separator with `-n` on non-contiguous matches

PR #16, opened 2026-06-22
feat(diff/patch): structured `diff --json` + GNU-fuzz `patch`

tobert:
## What

The diff/patch half of the [editing-for-agents](../blob/feat/diff-patch-improvements/docs/editing-for-agents.md) design pass. (The content-anchored `edit` builtin was **declined for kaish** — it isn't POSIX/shell-shaped and already lives in kaijutsu as an MCP tool.)

### `diff --json` — structured hunks
Emits `{old_file, new_file, hunks:[{old_start, old_lines, new_start, new_lines, changes:[{tag, content}]}]}` (`tag` ∈ `equal`/`delete`/`insert`, `content` newline-stripped for clean `jq`) via `OutputData::with_rich_json` — same mechanism as `grep --json`. Plain `diff` output is byte-for-byte unchanged. `diff -q --json` → `{old_file, new_file, differ:true}`.

### `patch` — GNU-style offset + fuzz (relaxes stricter-than-GNU behavior)
Previously `patch` hard-failed on *any* line drift — an attractive nuisance that loops agents. Now a pure, tested `apply_hunks`:
- locates each hunk by matching context near the header position (searching outward → **offset**),
- if exact match fails, ignores up to 2 lines of context per end (**fuzz**), searched within a bounded window of the header (a fuzzed match must land near where the patch says — prevents splicing into a distant coincidental match),
- rewrites only the verified span (fuzz never overwrites unverified lines),
- applies via a whole-file `PatchOp::Replace { offset:0, expected:Some(original) }` — TOCTOU-safe CAS, uniform across local/overlay backends,
- **preserves CRLF** line endings,
- reports offset/fuzz **loudly** (`Hunk #N succeeded at L (offset O lines)`), stays quiet on a clean apply,
- a hunk that matches nowhere **fails loud** (`Hunk #N FAILED`) and leaves the file untouched.

> [!WARNING]
> **BREAKING (exit-code contract):** a near-miss patch that previously exited non-zero (drift rejected) now applies and exits 0. Read the offset/fuzz report, not the exit code, to detect drift.

## Why content-anchored `edit` is *not* here
Reflex survey: agents reach for `patch`/`sed -i` by reflex. Research (2025–26): line-numbered unified diffs are the format LLMs generate *worst*. So `patch` stays a faithful workalike (relaxed, not amputated), and the reliable content-anchored path lives in the embedders (kaijutsu's hashline `edit`). Full rationale + rejected alternatives (incl. the silent "hijack patch → auto content-anchor" option) in `docs/editing-for-agents.md`.

## Tests
- `diff`: structured hunks, correct line ranges, pure-insertion line numbers, quiet `differ` object.
- `patch` (pure `apply_hunks`): clean / offset / fuzz / no-match-fails-loud / reverse / CRLF-preserved / bounded-fuzz-window.
- `patch` (execute-level): offset reported, failure loud + non-destructive.
- All gates green: `cargo test --all`, `cargo clippy --all --all-targets`. E2e smoke-tested through `./target/debug/kaish`.

## Review
High-effort multi-agent review run pre-commit. Fixed the confirmed correctness findings: CRLF collapse, unbounded fuzz search (→ bounded window), `--json` count-0 off-by-one, and the missing BREAKING marker. Deferred residuals (final-newline intent, file-creation `--- /dev/null` diffs) documented in `docs/issues.md` P4.


tobert:
## Review summary — Gemini Pro full-file pass + follow-on hardening

Ran a **Gemini Pro** (`gemini-pro-latest`, max-thinking batch) full-file review of `patch.rs` and `diff.rs`, prompted to "unwind" the code through error/edge paths. Every load-bearing finding was **verified against the real kernel path before acting** (the existing builtin tests call `.execute()` directly, bypassing arg-binding, so they didn't settle these). Verdicts below; fixes landed in the commits noted.

### `patch.rs`

| Finding | Verdict | Fix |
|---|---|---|
| `search_block` clamps `center` to `max_start` *before* computing the search distance, so the fuzz window is measured from the clamped position — a fuzzed hunk whose header points past EOF splices into a coincidental match far outside `FUZZ_SEARCH_WINDOW` | ✅ **Real**, reproduced (applied at `offset -999` on a 3-line file when it should fail loud) | `e6fabb2` |
| Redundant `search_block` call at `f=2` when context is 1+1 lines | ✅ Real, negligible | — |
| `--dry-run` re-reads pristine file for an exotic patch repeating one file across separate `---`/`+++` blocks | ✅ Real, exotic | — |
| Slice safety, non-destructive fuzz, CAS `expected`/`len` consistency, empty-source anchor math | Confirmed **sound** | — |

### `diff.rs`

| Finding | Verdict | Fix |
|---|---|---|
| *(SEVERE)* file extraction grabs `-C`'s value as `file1` (`diff -C 3 a b` opens "3") | ❌ **False** — `consume_flag_positionals` strips the value before execute; verified `diff -C 3 a b` works | — |
| Identical files return empty success with **no JSON object** — inconsistent shape for `--json` consumers | ✅ **Real**, verified | `e6fabb2` |
| Glued `-C3` triggers false "too few operands" rejection | ❌ **False** — verified `diff -C3 a b` works | — |
| `--json` drops missing-EOF-newline metadata the text path shows | ✅ Real | `59b2928` (`missing_newline` on each change object) |
| `line_no` ranges, op-span extraction, multi-line-value impossibility, text/JSON parity | Confirmed **sound** | — |

**Net from the review:** one real correctness/safety bug (the fuzz-window clamp — missed by the earlier pre-commit pass), plus two `--json` consistency gaps. The two "SEVERE" diff findings were both false alarms from the model not modeling kaish's arg-binder — worth noting for how much to trust this cast's execution-path reasoning without verification.

### Follow-on hardening (same branch, while in here)

- **`57eb5e3` — validator schema-aware arg binding.** Chasing a confusing `sed` error surfaced a live, *path-dependent* validation bug (`docs/issues.md`): the validator's arg-builder was schema-blind, so `sed -e1d -e2d FILE` fell back to parsing the **file path** as the sed program — a false `E006` that depended on the path's characters. The validation builder now binds glued/value/multi-consume short-flags exactly as execution does (shared helpers, can't drift). `sed -es/a/b/` now reports the accurate "s command requires delimiter".
- **`7695347` / `73f016c` — de-flaked two subprocess tests.** The cancellation and external-command-kill suites raced child fork/exec + PGID registration against fixed sleep/timeout windows under `--all` load. Both now wait for the observable readiness signal instead of guessing a duration.

**Gates:** `cargo test --all` and `cargo clippy --all --all-targets` green; every fix landed failing-first; e2e smoke-tested through `./target/debug/kaish`.

PR #24, opened 2026-06-24
fix(lexer): bare @, hyphenated numbers, and -1k lex as one bareword

tobert:
## What

Three everyday agent inputs used to fragment into adjacent tokens and trip the no-token-pasting guard — a *loud but wrong* parse error on ubiquitous text. All three are single contiguous words the user typed, not pasted tokens, so the fix is at the lexer (raw-slice tokens, so leading zeros survive), wired through the parser exactly where `NumberIdent` already flows.

| Input | Before | After |
|---|---|---|
| `user@host`, `a@b.com` | "unexpected character" | `Ident` (`@` is an ordinary word char) |
| `@scope/pkg`, `@0`, `@` | "unexpected character" | new `AtWord` token |
| `2024-01-02`, `10-20`, `1.5-2`, `cut -f 1-3` | adjacent-words parse error | new `DashNumWord` token |
| `find -size -1k`, `-30d` | parse error | `DashNumWord` |

A plain `2024`/`1.5`/`-1` still lexes as `Int`/`Float` — the hyphen form requires a `-segment` and the minus form an alpha after the digits, so they never overlap with `Int`/`Float`/`NumberIdent`.

## Decision: reverses the `tr -d 0-9` loud-error stopgap (intentional)

`tr -d 0-9` was previously made a *loud error* (pinned by `bareword_comma_tests::numeric_range_is_loud_not_silent`) only because `0-9` could fragment into `Int(0)`+`Int(-9)` and silently delete just `0`. A contiguous `0-9` the user typed is **one word**, so it now reaches `tr` verbatim and the range applies — bash/GNU parity. The pinning test was rewritten to assert the new behavior. `issues.md` listed `cut -f 1-3` as a bug and called the old error "misleading," so this is the intended direction.

**Not breaking:** every newly-valid input was a parse error before, so no previously-valid program changes meaning. The comma run (`echo 1,2,3`) stays a loud no-pasting error.

## Tests

- New kernel-routed `lexer_idiom_tests.rs` (13 cases incl. regression guards for `2024`/`-1`/comma-runs).
- Updated `lexer_tests.rs` (`-1a` → one `DashNumWord`) and `bareword_comma_tests.rs` (digit range now deletes digits).
- Full suite: 3716 pass, 0 fail. `clippy --all --all-targets` clean. No snapshot drift.

CHANGELOG / devlog / issues / LANGUAGE.md synced.


tobert:
## Review — deepseek (via kaibo), read-only against the branch

**Verdict: clean. No regressions; wiring complete; `@`-in-Ident safe.**

### Regex priority / longest-match interactions — all verified against existing tests
- `-1` → `Int` (DashNumWord regex-2 requires a trailing alpha; ShortFlag requires a letter after `-`). Pinned `lexer_tests.rs:589`.
- `1.5` → `Float` (DashNumWord regex-1's `(-…)+` group requires ≥1 `-segment`; logos won't match it zero times). Pinned `lexer_tests.rs:278-284`.
- `019dda1c-5b3f-7000` → `NumberIdent` (18-char match beats DashNumWord's 3-char prefix; longest-match wins at equal priority). Pinned `lexer_tests.rs:236`.
- `user@host` → one `Ident` (greedy from pos 0; AtWord never gets a mid-word start). `@0` → `AtWord`. `-1a` → `DashNumWord` (3 > 2 over `Int`).

### Wiring completeness — all token-consuming sites checked
`category()`, `Display`, `mergeable_text` (colon-merge), `glob_mergeable_text`, parser argv `select!`, case `pattern_part` — all present. Correctly **absent** from command-name position (matches `NumberIdent`). Not needed in `validator/walker.rs` (operates on parsed `Arg`, not raw tokens), `set.rs` option parser (reads post-parse `ToolArgs`), heredoc/interpolated-string scanners (char-level, no token kinds), redirect targets (reuse `primary_expr_parser`).

### `@` in the Ident class — no risk
`$@` is a fixed `#[token("$@")]` (always wins); `$foo@bar` → `SimpleVarRef("foo")` + `AtWord("@bar")`; `NAME=value`, function names, and the `set` keyword are unaffected; `user@host:8080` colon-merges into one word (tested).

### Non-blocking nits (not addressed; recorded)
- `1-` (trailing hyphen) still fragments into `Int(1)` + `MinusAlone` — pre-existing `MinusAlone` design tension, not a regression from this change.
- `set -o name=value` `option_value_str` doesn't accept `DashNumWord` — but `1-2` isn't a valid set option value anyway (loud parse error, not corruption).

### Aside found while adding a regression guard (pre-existing, orthogonal — left untouched)
`1..10` lexes as two `InvalidFloat` error tokens (`1.` + `.10`), **not** `Int`/`DotDot`/`Int`. This change doesn't touch it (DashNumWord requires a `-segment`); kaish has no numeric-range expansion by design.

Gates: 3716 tests pass · `clippy --all --all-targets` clean · no snapshot drift.

PR #25, opened 2026-06-24
fix: grep -c exit code, $() trailing-whitespace trim, jq division-by-zero

tobert:
Three small independent silent/surprise fixes from `docs/issues.md` (Amy batched them as the "small one-offs").

## 1. `grep -c` exits 1 on zero matches (GNU parity)
Was always exit 0. The count is still printed; over multiple files it exits 1 only when **no** file matched (a read error still overrides to exit 2). Both whole-buffer count paths set `code = 1`. (`-c` never takes the streaming path.)

## 2. `$()` strips only trailing newlines, not all trailing whitespace
The bare `Expr::CommandSubst` arms used `.trim_end()` — eating significant trailing spaces/tabs. Now `trim_end_matches(['\n', '\r'])`, matching the `"$(…)"` interpolation (`StringPart::CommandSubst`) and for-loop split paths. `x=$(printf 'a  ')` keeps the spaces (bash parity).

## 3. `jq '. / 0'` fails loudly instead of returning `null`
**Mechanism:** jaq evaluates `n/0` to a non-finite `Val::Float` (`inf`, or `NaN` for `0/0`); `val_to_json` did `from_f64(inf) → None → unwrap_or(Null)` — a silent-wrong null. New `has_nonfinite_float` (recurses arrays/objects) gates the result loop and returns a loud `jq runtime error` (exit 1). Modulo by zero already errored.

**Design note (worth a look):** this also errors on jq's `infinite`/`nan` *literals*, where real jq clamps `infinite` to max-f64 and renders `nan` as null. We can't distinguish a division-by-zero inf from an `infinite`-literal inf at serialization time — jaq (unlike real jq) doesn't error at the division itself — and a loud error beats a silent null per the no-silent-fallbacks / crash-over-corruption directive. `infinite`/`nan` literals are rare; accepted divergence. The separate jaq float-formatting quirk (`3.0` for `6/2`, `1e10`→`10000000000.0`) is untouched and stays in issues.md P3.

## Tests / gates
`correctness_oneoffs_tests.rs` — 8 kernel-routed (zero/some/multi-file grep counts, trailing-space-survives + newline-still-stripped, `/0` + `0/0` loud + finite division regression guard). Full suite 3724 pass, 0 fail. `clippy --all --all-targets` clean. Verified in the real binary.

CHANGELOG / issues / devlog synced (issues.md: the readlink half of the grep bullet was already fixed in the P2 batch — removed as stale).


tobert:
## Review status

**kaibo is unavailable for this PR** — the server errors on its `[orientation] full_list_max_files (256)` cap (this repo is 464 files) before it can explore, for every cast (it's a server-config limit, not cast-specific). No external review obtained.

### Self-review (the checks a reviewer would run)

**grep -c** — `-c` cannot reach either streaming path (both gated by `is_simple = !count_only && …`), so it always hits a whole-buffer count path; both are fixed (single-buffer + `grep_multiple_files`). Recursive directory grep funnels through `grep_multiple_files`, covered. In multi-file, the later `error_text` override (`code = 2` on a read error) runs *after* my `code = 1` and correctly wins. `-cq` → quiet branch precedes count (GNU-consistent). `-cv` counts inverted matches, exits 1 if none — sane.

**$() trim** — refined in the follow-up commit (`890d101`): bare `$(cmd)` and quoted `"$(cmd)"` are the same operation and must trim identically, so I aligned both bare arms to `trim_end_matches('\n')` — the *exact* trim the interpolation path uses — rather than the `['\n','\r']` I first used (the for-loop split path). Otherwise `x=$(printf 'hi\r')` and `x="$(printf 'hi\r')"` would diverge on a trailing `\r`. The `out_bytes` and `.data` branches are untouched (correct — only the two string branches over-trimmed).

**jq non-finite** — the `has_nonfinite_float` guard runs before the format branch, so it covers raw/compact/json alike. `Val::Num` (string-backed bignum) is never non-finite (JSON input can't carry inf/nan; arithmetic yields `Val::Float`), so not checking it is correct. The one accepted divergence — erroring on the rare `infinite`/`nan` literals — is documented in the PR body and devlog.

Full suite 3724 pass · `clippy --all --all-targets` clean · all three verified in the binary.

tobert:
## kaibo review (deepseek) — outcomes

Reviewed at `crates/kaish-kernel/src` (137 files — under kaibo's 256-file orientation cap; the repo root's 464 files tripped it). **Verdict: fixes are sound.** Two non-blocking residuals raised; both now resolved.

### Confirmed correct
- **grep -c** — `-c` is walled off from both streaming paths by the `is_simple` (`!count_only`) gate, so it always reaches a whole-buffer count path (single-buffer stdin/file, multi-file, and `-r` recursive → all covered). The multi-file `error_text` override correctly turns `code=1`→`2` on a read error, while the `quiet && matches>0` exception keeps POSIX's exit-0-on-match. `-cq` (quiet wins) and `-cv` (counts inverted matches) both sane.
- **$() trim** — bare `$(cmd)` now uses the exact `trim_end_matches('\n')` the quoted `"$(cmd)"` interpolation path uses (`kernel.rs:3835`); the `out_bytes`/`.data`/flat-node branches correctly don't trim.
- **jq non-finite** — the `has_nonfinite_float` guard runs before all three format modes *and* before `values.push`, so the structured `.data`/`--json`/for-loop path is protected too, not just text.

### Residuals — addressed
1. **Sync evaluator still used `.trim_end()`** (`interpreter/eval.rs:798`, `result_to_value`). Dead in production (runs under `NoOpExecutor`, which errors first), but a trap for a future non-async embedder. **Fixed** in `32a7a4b` → `trim_end_matches('\n')`.
2. **`Val::Num` not checked in `has_nonfinite_float`** (defense-in-depth suggestion). **Declined, with reasoning:** a non-finite `Val::Num` can't occur — overflowing inputs like `1e400` are rejected at JSON parse ("number out of range"), and arithmetic produces `Val::Float`, not `Num`. Even in the impossible case, `Num` renders as a visible string `"inf"`, *not* the silent `null` this fix targets — so it's outside the bug class and would only add a per-value parse.

Pre-existing (not from this PR, not addressed): multi-file `grep -c` prints a single total, not GNU's per-file `file:count` lines.

Gates after the follow-up: full suite **3724** pass, 0 fail · `clippy --all --all-targets` clean.

PR #29, opened 2026-06-25
remove: drop the `test` builtin and `[` command

tobert:
## What

Removes the `test` builtin and the `[` command. Conditionals go through `[[ … ]]` only.

## Why

The removed builtins never worked end-to-end through the parser:

- **`[` (Bracket) was dead code** — it can't be a command name (`[` lexes as `LBracket`, which `command_parser`'s `command_name` doesn't accept).
- **`test`'s comparison half never routed** — `=`/`!=`/`<`/`>` lex as operator tokens the argv parser rejects in argument position, so `test a = b`, `test 5 -eq 5`, etc. failed to **parse**. Only unary `test -z`/`test -f` reached the builtin.

Their unit tests passed only because they called `.execute()` directly with pre-built `ToolArgs`, bypassing lex/parse — the exact gotcha CLAUDE.md warns about ("test builtins through `kernel.execute(...)`").

Rather than the delicate, broad-blast-radius parser change to accept those operator tokens as barewords (without breaking assignment `x=y`, glob char-classes `ls [ab]*`, or `[[ ]]`), we keep `[[ … ]]` as the single test form. **`[[ … ]]` is real grammar the validator understands**, so a malformed test / unknown operator / unquoted expansion is caught *before* runtime — where `test`/`[` hide their operators as runtime string arguments, the late-failure footgun kaish exists to remove.

## Changes

- Delete `test_builtin.rs` (`Test` + `Bracket`); drop their registration in `mod.rs`.
- Fix the `limits.md` falsehood (it claimed `test` worked as a fallback — it didn't) and document the deliberate absence + validation rationale.
- Convert the sandbox host-probe test and the json-sweep `--json` case to `[[ ]]`; drop the resolved `issues.md` entry.
- CHANGELOG: **BREAKING** removal. devlog: standing decision (don't re-litigate).

## Notes

In a full REPL build (subprocess + seeded PATH), `test`/`[` now resolve to the host's `/usr/bin/test` like any other external command — bash behaves the same. In a sandboxed/no-subprocess embedder they're cleanly "command not found." Either way kaish provides no `test`, and the validator can't pre-check an opaque external command — consistent with the rationale above.

## Gate

- `cargo test --all` — green
- `cargo clippy --all --all-targets` — clean
- Smoke: `[[ -n hello ]] && echo YES` → `YES` (exit 0)


tobert:
## Update: gemini-batch review (via kaibo) + follow-ups

Ran a `gemini-batch` review through kaibo. It confirmed the rationale is sound and the drift guard/test conversions are correct, and caught three real items — all addressed in `8394362`:

1. **Silent VFS-escape footgun (the important one):** with the builtin gone, a bare `test -f file` in a subprocess build falls through to host `/usr/bin/test`, which evaluates against the **real** filesystem, bypassing the VFS/overlay — a silent wrong boolean into `if`/`&&`. **Fix:** new validator advisory **W006** (`PosixTestCommand`) steers `test` → `[[ … ]]` and *still runs* (warn-don't-reject). This is the first *agent-surfaced* validation warning — warnings were trace-only (every external command fires `UndefinedCommand`), so a code now opts into surfacing via `IssueCode::surfaces_to_agent`, the seam for the broader did-you-mean pass. Also dropped `test`/`[`/`[[` from `is_special_command` so `test` actually reaches the advisory.
2. **Stale docs:** three "use the `test` builtin" references in `docs/LANGUAGE.md` fixed.
3. **Dead code:** removed the now-unreachable `Skip { name: "[" }` from the json-sweep guard.

New tests: walker unit test (emits W006, not `UndefinedCommand`) + two kernel-routed tests (W006 surfaces & still runs; the generic undefined-command warning stays trace-only). `cargo test --all` green, `cargo clippy --all --all-targets` clean.

PR #30, opened 2026-06-25
fix: write binary content from $() verbatim, not the placeholder

tobert:
## What

`write FILE $(producer)` (and `producer | write FILE`) now persists the raw bytes of a `Value::Bytes` verbatim instead of stringifying it to the literal text `[binary: N bytes]` and writing **that** to the file — silent corruption.

First PR of the **0.9.2 write-model bundle**.

## Why

`write`'s content path ran every value through a local `value_to_string` whose `Bytes` arm produced `[binary: N bytes]`. So `write out.bin $(cat some.bin)` wrote the 16-ish-char placeholder, not the bytes. The stdin branch also used `read_stdin_to_text` (lossy/erroring on binary). Crash-or-correct is the standard; we never silently corrupt.

## Fix

Read `write`'s content as raw bytes from every source:
- a new `value_to_bytes` helper writes `Bytes` verbatim; other variants use their lossless string form,
- stdin uses `read_stdin_to_bytes` (pipe-preferred), matching `tee`.

## Scope note

The issue named `write`/`tee`, but `tee`/`patch` were verified **already safe** and left untouched: `tee` reads stdin as raw bytes (its `value_to_string` is for the *path* operand, not content), and `patch`'s content is a text diff. `tee` is pinned by a regression test so it stays that way.

## Tests

Kernel-routed (`binary_write_tests.rs`), invalid-UTF-8 fixture so the producer yields `Value::Bytes`:
- `write_positional_cmdsubst_preserves_bytes` — reproduces **red** against the pre-fix stringify
- `write_pipe_stdin_preserves_bytes`
- `tee_preserves_bytes_through_to_file` (pin)

Also smoke-verified byte-identical 1 KiB `/dev/urandom` round-trips for both paths.

## Gate

`cargo test --all` green · `cargo clippy --all --all-targets` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,14 @@ breaking entries are marked **BREAKING**.
 
 ## [Unreleased]
 
+### Fixed
+- **`write` no longer corrupts binary content from `$(…)`.** `write FILE $(producer)`
+  (and `producer | write FILE`) now persists the raw bytes of a `Value::Bytes`
+  verbatim instead of stringifying it to the `[binary: N bytes]` placeholder — that
+  placeholder reaching a file was silent corruption. Content from a positional,
+  `--content`, or stdin is all read as bytes. (`tee` already read stdin as raw bytes
+  and is unaffected; pinned by a regression test.)
+
 ### Added
 - **Validator advisory `W006` steers `test` to `[[ … ]]`.** Using a bare `test` as a
   command now surfaces a one-time stderr note (`use [[ … ]] …`) and still runs. A
--- a/crates/kaish-kernel/src/tools/builtin/write.rs
+++ b/crates/kaish-kernel/src/tools/builtin/write.rs
@@ -66,38 +66,42 @@ impl Tool for Write {
             None => return ExecResult::failure(1, "write: missing path argument"),
         };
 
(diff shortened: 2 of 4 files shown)


tobert:
## gemini-pro batch review (via kaibo) — addressed

Ran a real offline **batch** review this time (`batch_submit`, model `gemini-pro-latest`, files attached). It confirmed:
- `value_to_bytes` is correct for every `Value` variant (100% fidelity vs the old string form);
- `tee` is immune (reads raw bytes; its `value_to_string` is path-only) and `patch` is immune (text tool, loud-errors on a binary target).

It flagged the `read_stdin_to_bytes` empty-stream question and three test gaps. Verified + closed in `1b47237`:
- **Empty stream is correct, not a bug** — `read_stdin_to_bytes` returns `None` only when stdin is truly absent; an empty pipe is `Some(vec![])`. Pinned by `write_empty_pipe_creates_empty_file` (`printf '' | write f` → empty file, exit 0).
- Added `write_named_content_flag_preserves_bytes` (`--content $(cat bin)` — the named branch) and `tee_preserves_bytes_to_multiple_sinks` (`tee a.bin b.bin` — buffer not drained after the first sink).

All 6 tests green.

PR #31, opened 2026-06-25
feat: gate tee overwrites through latch + trash (write-model)

tobert:
## What

`tee` now honors `set -o latch` and `set -o trash` on a **truncating overwrite**, the way `rm` gates a delete. Second PR of the **0.9.2 write-model bundle**; `patch` and `sed -i` reuse the same infra next.

## Why

`rm` has gated destructive deletes (latch + trash-on-delete) for a while, but the *overwriting* builtins bypassed those rails — an agent could silently truncate a file with no confirmation and no recoverable prior copy. This closes that for `tee`.

## Shared infra (reused by patch/sed next)

- **`decide_mutation_action → {Proceed, TrashFirst, Latch}`** — a pure decision table mirroring `rm`'s trash-over-latch priority (unit-tested). New file / `tee -a` append / `/tmp`,`/v` real paths → `Proceed`. An overlay/in-memory file has no real path, so it's *not* excluded and stays gated (Amy: protections are about agent-op safety, not just real-FS data).
- **`ExecContext::gate_overwrites(command, targets, confirm)`** — stats each target, issues **one** latch nonce scoping the whole batch (exit 2 until `--confirm=<nonce>`), and snapshots every trash-first target before any write. `snapshot_for_overwrite` moves a real file to trash, or captures an overlay file's current bytes.
- **`TrashBackend::trash_bytes(original_path, bytes)`** — the overlay byte-snapshot entry point you approved. `SystemTrash` materializes the bytes to a staging file (basename preserved) and trashes it; what's recoverable is the bytes, not the location.

## tee

Added `--confirm`; calls `gate_overwrites` before writing. Append and new files don't gate. **Both gates are off by default, so default `tee` is unchanged.**

## Tests

- 6 decision-table unit tests (`tools::context::tests`).
- 4 kernel-routed tee tests (`latch_trash_tests.rs`): trash-first real-file snapshot, latch + `--confirm` round-trip, new-file/append no-gate, and the **overlay `trash_bytes` path** via an in-memory `/v` mount.
- Help fragment + regenerated `syntax.md` note the broadened latch/trash scope.

## Scope / known limitation

`SystemTrash::trash_bytes` stages the snapshot under the system temp dir then trashes it — restore lands at the staging path, so the *bytes* are recoverable but not the original location. Fine for the overlay/in-memory case (rare); real-FS overwrites use the solid `trash(real_path)` move. Flagging for review.

## Gate

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,12 +23,12 @@ breaking entries are marked **BREAKING**.
   (pure-Rust `infer`, no C deps) — `detect` for magic-only, `classify` for the
   text-aware path — so embedders can classify bytes the same way the shell does.
 - **`tee` honors `set -o latch` / `set -o trash` on a truncating overwrite**, like
-  `rm` gates a delete. Overwriting an existing file under `trash` snapshots its prior
-  content first (the real file is moved to trash, or — for an overlay/in-memory file —
-  its bytes are captured via the new `TrashBackend::trash_bytes`); under `latch` it
-  needs `tee --confirm=<nonce>` (exit 2 until confirmed, one nonce scoping all files).
-  A new file or `tee -a` append never gates. Both gates are off by default, so default
-  `tee` is unchanged. (`patch`/`sed -i` adopt the same gate next.)
+  `rm` gates a delete. Overwriting an existing file under `trash` first copies its
+  prior content to trash (recoverable, via the new `TrashBackend::trash_bytes`),
+  leaving the file in place for the new content; under `latch` it needs
+  `tee --confirm=<nonce>` (exit 2 until confirmed, one nonce scoping all files). A new
+  file or `tee -a` append never gates. Both gates are off by default, so default `tee`
+  is unchanged. (`patch`/`sed -i` adopt the same gate next.)
 - **Validator advisory `W006` steers `test` to `[[ … ]]`.** Using a bare `test` as a
   command now surfaces a one-time stderr note (`use [[ … ]] …`) and still runs. A
   path-qualified form (`/usr/bin/test`, `./test`) is left alone — it's an explicit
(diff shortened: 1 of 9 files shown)


tobert:
## gemini-pro review (write-model stack) — addressed in `88405d3`

Reviewed the shared overwrite-gate rails via a gemini-pro batch. Findings on this PR and their resolution:

- **High — unbounded staging-dir leak in `SystemTrash::trash_bytes`.** Each snapshot created `/tmp/kaish-snapshot-<nanos>/`, trashed the file inside, and never removed the now-empty dir. Fixed: `remove_dir` after trashing.
- **High/Medium — staging in `/tmp`.** A tmpfs `/tmp` routes the trash to `/tmp/.Trash-$UID` and loses the snapshot on reboot, and restore lands at an opaque path. Fixed: stage under the persistent data dir (`~/.local/share/kaish/trash-staging`) so the move stays single-fs and lands in the durable `~/.local/share/Trash`.
- **Medium — nanosecond staging-path collision** under parallel agents / tight loops. Fixed: mix a process-global `AtomicU64` counter into the dir name.
- **Medium — tee cloned the full input buffer per non-append operand.** Fixed: write the borrowed input directly on overwrite; only append does a read-modify-write.
- **Low — tee kept only the last per-file error.** Fixed: accumulate and report every failure.

Also reworked `gate_overwrites` to take a `confirm_hint` closure `(nonce, joined)` and to **dedup targets by resolved path centrally** — this is what lets `sed -i` reinject its flags into the latch hint (#34) and protects every caller from N-duplicate snapshots/prompts (#32).

Tradeoffs left as-is per the review's own assessment (inherent without a read-and-snapshot VFS primitive): the snapshot↔CAS double-read TOCTOU gap and upfront-snapshot orphans on early failure.

PR #32, opened 2026-06-25
feat: gate patch overwrites through latch + trash (write-model)

tobert:
## What

Wires `patch` into the shared write-model gate from #31, so an agent can't silently rewrite a file under `set -o latch` / `set -o trash` with no confirmation and no recoverable prior copy. **Third PR of the 0.9.2 write-model bundle**; `sed -i` is the last.

**Stacked on #31** (`feat/mutation-write-gate`) — base is that branch, not main. Review #31 first.

## How

`patch` always truncating-overwrites an existing file, so each target is a non-append overwrite. After parsing + grouping hunks, it calls `gate_overwrites("patch", targets, --confirm)` with **one nonce scoping every file the diff touches**, skipped for `--dry-run` (which never writes).

The copy-bytes snapshot from #31 leaves the file **in place**, so patch's existing read → apply → compare-and-swap write path is unchanged — that's exactly why #31 switched the snapshot from move to copy (a move would have yanked the file out from under patch's read).

Added `--confirm` to the clap layer.

## Tests

Kernel-routed (`latch_trash_tests.rs`):
- `patch_overwrite_under_trash_snapshots_prior_bytes` — prior content captured before the patch lands.
- `patch_overwrite_under_latch_requires_confirm` — exit 2 + `--confirm` round-trip; file untouched until confirmed.
- `patch_dry_run_does_not_gate` — `--dry-run` never gates.

Help fragment + regenerated `syntax.md` note `patch` joining the gate.

## Gate

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,13 +22,14 @@ breaking entries are marked **BREAKING**.
   `filetype` module
   (pure-Rust `infer`, no C deps) — `detect` for magic-only, `classify` for the
   text-aware path — so embedders can classify bytes the same way the shell does.
-- **`tee` honors `set -o latch` / `set -o trash` on a truncating overwrite**, like
-  `rm` gates a delete. Overwriting an existing file under `trash` first copies its
-  prior content to trash (recoverable, via the new `TrashBackend::trash_bytes`),
-  leaving the file in place for the new content; under `latch` it needs
-  `tee --confirm=<nonce>` (exit 2 until confirmed, one nonce scoping all files). A new
-  file or `tee -a` append never gates. Both gates are off by default, so default `tee`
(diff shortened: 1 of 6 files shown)


tobert:
## gemini-pro review — addressed in `811aac6`

- **Medium — duplicate targets with `--file` + multi-file diff.** An explicit target listed once per hunk-group would snapshot to trash N times and list the file N times in the latch prompt. Resolved by the **central dedup** now in `gate_overwrites` (#31); patch just adopts the new `confirm_hint` closure. Added tests: a two-group diff against one explicit target snapshots prior bytes exactly once, and the latch prompt never repeats the deduped target.
- **Validated correct** by the review: `--dry-run` bypass, non-existent-target handling, and the upfront-batch snapshot tradeoff (single latch prompt vs. mid-operation).

PR #34, opened 2026-06-25
feat: sed -i in-place edit, gated through latch + trash

tobert:
## What

`sed -i` edits one or more file operands in place instead of streaming to stdout — the strongest remaining sed ergonomic gap (both sed-panel models reached for `sed -i 's/…/…/' file`). **Last PR of the 0.9.2 write-model bundle.**

**Stacked on #32** (`feat/patch-write-gate`) → which is stacked on #31. Review the stack in order.

## How

`sed -i` is always a truncating overwrite, so it routes through the shared `gate_overwrites` from #31 (latch + trash, one nonce per command). Added `--confirm`. No file operands is a loud error (editing a stream in place is meaningless). Streaming mode is untouched; the 78 existing sed unit tests still pass.

## Scope cut: `-i.bak` is NOT supported (architectural, flagged for your call)

Amy's spec wanted the GNU glued backup suffix `-i.bak`. It turns out kaish can't model it post-lexing:
- the **lexer** splits `-i.bak` at the dot (dot-prefixed barewords → `-i` + `.bak`), and
- the kernel **value-flag binder** consumes the *next* token greedily, so declaring `-i` as a value flag makes `sed -i 's/…/…/' f` eat the expression as the suffix.

Neither glued nor spaced optional-suffix survives. Recovering it needs lexer/binder work (or a bespoke pre-clap argv pass à la `set.rs`). Since the **trash snapshot already keeps a recoverable prior copy**, the explicit `.bak` is convenience, not safety — so I shipped `-i` without it and filed `-i.bak` + the family-wide atomicity question (overwrite-in-place vs write-temp-rename) in `issues.md`. Happy to revisit `-i.bak` as its own lexer PR if you want it.

## Tests

Kernel-routed (`latch_trash_tests.rs`): trash snapshots prior bytes before the in-place write; latch gates with exit 2 + `--confirm` round-trip; no-operands loud error. Docs synced (CHANGELOG, LANGUAGE.md sed scope, help fragment + regenerated `syntax.md`, issues.md consolidated to DONE + residuals).

## Gate

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,9 +25,12 @@ breaking entries are marked **BREAKING**.
 - **`sed -i` edits files in place** instead of streaming to stdout, across one or
   more file operands. It is always a truncating overwrite, so it routes through the
   same latch/trash gate as `tee`/`patch` below. With no file operands it's a loud
-  error (editing a stream in place is meaningless). The GNU glued backup suffix
-  (`-i.bak`) is **not yet supported** — kaish's lexer splits `-i.bak` at the dot; the
-  trash snapshot under `set -o trash` already keeps a recoverable prior copy.
+  error (editing a stream in place is meaningless). Each file is written with a
+  whole-file compare-and-swap (like `patch`), so a concurrent change between read and
+  write is a loud conflict, never a silent clobber; a per-file failure doesn't abort
+  the batch and every error is reported. The GNU glued backup suffix (`-i.bak`) is
+  **not yet supported** — kaish's lexer splits `-i.bak` at the dot; the trash snapshot
+  under `set -o trash` already keeps a recoverable prior copy.
 - **`tee`, `patch`, and `sed -i` honor `set -o latch` / `set -o trash` on a truncating
   overwrite**, like `rm` gates a delete. Overwriting an existing file under `trash`
   first copies its prior content to trash (recoverable, via the new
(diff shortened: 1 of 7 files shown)


tobert:
## gemini-pro review — addressed in `60af159`

- **Critical — latch hint dropped `-i` and the expression → agent hangs.** `gate_overwrites` hardcoded `{command} --confirm=… {paths}`, so sed advertised `sed --confirm="NONCE" file` — which reads `file` as the expression and then blocks on stdin forever. Fixed: the gate now takes a `confirm_hint` closure and sed rebuilds a runnable `-i [-n] -e '<expr>'…` hint. Test runs the advertised hint **verbatim** and requires it to edit the file.
- **High — missing `flagify_bool_named`.** A structured `{"in-place": true}` call rendered `--in-place=true`, which clap's `SetTrue` rejects, breaking structural tool calls. Fixed to match patch (`-i.bak` is still left for clap to reject as the unsupported glued-suffix form).
- **Medium — TOCTOU.** Blind `WriteMode::Overwrite` silently destroyed a concurrent edit. Fixed: whole-file compare-and-swap (`PatchOp::Replace` with `expected`), matching patch.
- **Low — only last per-file error reported.** Fixed: report every error.

Added in-place tests: runnable hint, multi-file + `-e`, and continue-past-per-file-error. CHANGELOG notes the CAS-safety.

All four of the review's specific questions (file_pos/-i routing, continue-on-error, parsed-vs-raw flag, test adequacy) are covered.

PR #35, opened 2026-06-26
fix: correctness polish bundle (tilde file-tests, <<- tabs, export scope, jq numbers)

tobert:
A bundle of small, independent correctness fixes — each its own commit, TDD'd
(failing test first, verified red against a fix-reverted build where practical),
with CHANGELOG + issues.md updated per fix.

## Fixes

1. **`[[ ]]` file tests expand `~`.** `[[ -f ~/x ]]` (and `-d`/`-e`/`-r`/`-w`/`-x`)
   was always false — `eval_test_async` stat'd the literal `~/x`. Routes through
   `apply_tilde_expansion(scope_home())` like argv positionals; hermetic kernels
   keep `~` literal. (`file_test_tilde_tests.rs`)

2. **`<<-` strips source tabs, not `$var` tabs.** POSIX `<<-` strips leading tabs
   from each *source* line before parameter expansion, so a tab from a `$var` value
   at line start must survive (bash parity). New `HeredocAssembler` assembles the
   body part-by-part with a line-start state machine; wired into all three
   interpolated-heredoc paths (sync eval, async kernel, sync pipeline redirect).
   (`heredoc_tests.rs`)

3. **`export NAME=VALUE` in a function publishes to shared scope.** It wrote the
   function frame, so the value vanished on return (the export *mark* survived,
   leaving an exported name with no value). New `Scope::set_exported_global`
   (global set + mark); `set_exported` stays for the intentional frame-scoped
   overlays (`execute_with_vars`, `FOO=bar cmd`). (`export_scope_tests.rs`)

4. **`jq` renders integral numbers without `.0`.** `6/2`→`3`, `1e10`→`10000000000`;
   `5/2`→`2.5` and `1e100`→`1e+100` unchanged. `jq_float_to_json` + `num_str_to_json`
   on both the `-c`/json and `-r` paths. (`jq_tests.rs`)

Closes the issues.md P2 "Interpreter trio" entirely (items 1–3). Two further jaq
items are deferred **with root causes recorded** in issues.md: `keys_unsorted` sorts
because `serde_json` lacks `preserve_order` (insertion order is lost at parse — a
workspace-wide IndexMap change), and `null | .a` erroring is jaq-core field-access
semantics (needs a jaq bump).

## Gates
`cargo test --all` clean, `clippy --all --all-targets` zero warnings, heredoc
bash-compat leg green.


tobert:
## Review (retrospective — merged)

Reviewed by a local general-purpose subagent against the actual sources before merge. Verdict: clean, no real bugs in the new logic.

- `HeredocAssembler` `<<-` semantics correct: consecutive leading tabs stripped, interpolations terminate the leading-tab run (even when empty), newlines reset only via real literal parts; all three call sites bucket `Literal` vs interpolation correctly.
- `Scope::set_exported_global` matches plain-assignment frame semantics; frame-scoped overlays (`execute_with_vars`, `EnvScoped`) still use `set_exported` — unchanged.
- `jq_float_to_json`/`num_str_to_json` (since superseded by #36): 2^53 bound, big-int preservation, negatives all sound.

One review finding actioned pre-merge: the tilde negative test wasn't differential (a literal `~/nope` is absent for the wrong reason) — replaced with a HOME≠cwd case, verified red against a fix-reverted build.

Gates green: `cargo test --all` + `clippy --all --all-targets`, heredoc bash-compat leg green.

PR #36, opened 2026-06-26
chore: upgrade jaq to v3 — fixes null-index, object ordering, big ints

tobert:
Upgrades the native jq engine to **jaq-core 3 / jaq-std 3 / jaq-json 2**, which
fixes both jaq items previously deferred in `issues.md` (root causes were
dependency-level), plus a bonus. I confirmed the fixes in the crate source
*before* taking on the migration.

## What this fixes (user-visible)

- **`null | .a` → `null`** (was `cannot use null as iterable`) — jaq-json 2's
  `index_opt` maps `null → None → null`, matching real jq.
- **`keys_unsorted` and object output preserve insertion order** — jaq-json 2's
  `Val::Obj` is an `IndexMap`. To keep that order across the input parse, this also
  enables **`serde_json`'s `preserve_order` workspace-wide**, so *all* JSON object
  output (`--json`, `.data`) is now insertion-ordered (jq parity) instead of
  key-sorted. `jq keys` (sorted) is unchanged.
- **Large integers are exact** — `9999999999999999999999` round-trips verbatim
  instead of degrading to `1e+22` (bignum-backed `Num`).
- Number canonicalization preserved from the prior fix: `6/2` → `3`, `1e10` →
  `10000000000`, `5/2` → `2.5`.

## Why it's a real migration (not a version bump)

jaq 3.x restructures the `Val` enum (numbers unified into a bignum `Num`, strings
are byte-strings, objects are `IndexMap`) and the run API (`Compiler`,
`Ctx::<JustLut<Val>>::new(&filter.lut, Vars::new(...))`, `filter.id.run(...)` +
`unwrap_valr`). Notable choices:

- Output now goes through **jaq's own writer** (`jaq_json::write`) for compact/pretty
  rendering, with a `canonicalize_numbers` pass that turns integral floats into
  integers (jaq's writer prints `3.0`/`1e10` otherwise).
- `val_to_json` (for `.data`) round-trips jaq's compact rendering back through
  `serde_json`, so `.data` number formatting and key order match stdout exactly.
- Division-by-zero **still** yields a non-finite float in jaq 3.x, so the loud
  `has_nonfinite_float` guard stays (adapted to `Val::Num(Num::Float)`).

## Tests / gates

`jq_tests.rs` gains null-index, `keys_unsorted`/object-order, `keys`-still-sorted,
big-int, and pretty-print cases (29 jq tests total). `cargo test --all` and
`clippy --all --all-targets` are green **with `preserve_order` on** — no test
depended on sorted keys.

## Note on residual

`1e100` renders as `1e100` where real jq prints `1e+100` — a cosmetic difference on
an out-of-exact-range float; not worth special-casing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,15 @@ breaking entries are marked **BREAKING**.
   otherwise resolve to an external `/usr/bin/test` that evaluates against the real
   host filesystem, bypassing the VFS/overlay (a silent wrong boolean into `if`/`&&`).
 
+### Changed
+- **`jq` object keys now preserve insertion order** instead of being sorted
+  alphabetically — `echo '{"b":1,"a":2}' | jq .` emits `{"b":1,"a":2}`, matching
+  real jq. This comes from upgrading the native jq engine to jaq-core 3 / jaq-json 2
+  and enabling `serde_json`'s `preserve_order` workspace-wide, so any tool emitting
+  JSON objects (`--json` output, structured `.data`) now keeps source/insertion
+  order rather than sorting keys. `jq keys` (sorted) is unchanged; `keys_unsorted`
+  now correctly reports source order.
+
 ### Removed
 - **BREAKING: the `test` builtin and `[` command are gone.** Use `[[ … ]]`, the one
   supported test form. The removed builtins never worked end-to-end through the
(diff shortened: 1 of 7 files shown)


tobert:
## Reviews

Two independent review passes, both followed up on.

**1. Local subagent review** (general-purpose, against jaq-json 2.0.1 / jaq-core 3.1.0 sources):
- **32-bit `isize` saturation** in `canonical_integral` (`1e10`→`2147483647` on wasm32) — fixed (`Num::from_integral(f as i64)`), and `cargo build -p kaish-wasi --target wasm32-wasip1` confirmed green.
- **big integers lossy in `.data`** (stdout exact, `.data` f64-limited >2^63) — documented as a residual (not a regression; needs `arbitrary_precision`, a separate broad change).
- `json_to_val` number else-arm now panics rather than silently returning `Val::Null` (crash-over-corruption).
- `has_nonfinite_float`, `preserve_order` blast radius, run/compile API, zero-snapshot-impact — verified clean.

**2. kaibo review** (`cast=deepseek`), 6 findings triaged against the jq 1.8.1 oracle:
- ✅ **2^53 boundary** (`<` → `<=`): `pow(2;53)` now prints `9007199254740992` (was `…992.0`). Fixed + test.
- ✅ **object-key canonicalization**: `{(6/2):"x"}` → `{3:"x"}` (jaq tolerates numeric keys; render them consistently). Fixed.
- ❌ duplicate `--arg x 1 --arg x 2` / interleaved `--arg`+`--argjson` "resolve first-match" — **false positive**, verified last-wins (`"2"`/`2`) matching jq.
- ❌ `has_nonfinite_float` "misses `Num::Dec`" — **non-reachable**: `nan`/`infinite`/`÷0` are always `Num::Float`, caught loudly.
- ❌ `val_to_json` String fallback — only fires on non-UTF-8 byte strings, where stdout already lossy-decodes to the same; documented.

## Gates
`cargo test --all` (incl. 30 jq tests) + `clippy --all --all-targets` green with `preserve_order` on; wasm32-wasip1 builds clean.

PR #38, opened 2026-06-28
feat: rg-style file-type filtering for grep + glob (--ftype, --max-count, --hidden)

tobert:
Ports the still-useful ripgrep-only filtering onto kaish's two *modern* search
builtins — **grep** and **glob** — reusing the `ignore::types` machinery the
walker already wrapped (it was live but dormant: no builtin populated
`WalkOptions.types`). `find` stays POSIX, untouched. Driver: **kaibo**, whose
hot path is "grep this codebase" — `grep --ftype rust` and `--max-count` are the
ergonomics that make it fast and precise.

## What's new

| builtin | flags | notes |
|---|---|---|
| **grep** | `--ftype`, `--ftype-not`, `--ftype-list`, `--hidden`, `--max-count` | file-type filter on the `-r` walk; `--max-count` early-stops the chunk reader (big-file win) |
| **glob** | `--ftype`, `--ftype-not`, `--ftype-list` | keeps `-t`/`--type` = entry-kind (file/dir, fd-style); the two axes compose |
| **find** | — | unchanged, POSIX |

## Design decisions (from the conversation)

- **`--ftype` is the kaish-wide file-type-filter standard**, not rg-native `-t`.
  `glob` already owns `-t`/`--type` for *entry kind*; file-type is a different
  axis and gets one consistent name across search builtins. Recorded in
  `docs/NAMING.md`.
- **New flags are long-only, no shorts** — sidesteps GNU grep's `-T`/`--initial-tab`
  and `-m` muscle memory; kaish borrows the useful long names, not the short surface.
- **Unknown type name is a loud exit-2 error**, never a silent empty match.
- **`--no-ignore` deliberately deferred** — won't ship a flag whose `Enforced`-scope
  override semantics aren't designed yet (tracked in `docs/issues.md`).
- Shared `kaish-glob::build_file_types` + `tools/builtin::read_repeatable_strings`
  so grep and glob can't drift.

## `--max-count` (GNU semantics, per file)

Threaded through all four match paths — streaming-stdin, single-file chunked
scanner, whole-buffer, and multi-file. Streaming paths halt the chunk reader once
the cap is hit (the kaibo big-file win). `--max-count 0` matches nothing (exit 1).
The whole-buffer path lets trailing after-context of the Nth match through
(`grep -m N -A k`).

## Review

DeepSeek reviewed via kaibo (Gemini was provider-down). Verdict: design sound
across the board; found one rare "loud-in-the-wrong-direction" bug — a
`--max-count` stop at a 256 KiB boundary splitting a UTF-8 multibyte char made
`finish()` flag the truncated carry as binary (spurious exit-2). **Fixed**
(`finish()` now guards on `hit_limit`) with a regression test verified red-then-green.

## Tests

24 new: 5 `kaish-glob::file_types` unit + 12 grep + 6 glob kernel-routed
(`*_search_features_tests.rs`) + 1 scanner regression. Covers repeatable
`Json(Array)` accumulation, per-file `--max-count` across all paths + the
truncated-carry case, loud unknown-type, `--hidden` dotfile reach, the glob `-t`
entry-kind regression, and the dirs-no-op interaction. `cargo test --all` and
`cargo clippy --all --all-targets` clean throughout.

`docs/search-features-port.md` is the design + punch-list (transient — delete on
the follow-up once it's served its purpose).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ breaking entries are marked **BREAKING**.
   TYPE→globs table (works with `--json`); no pattern required.
 - **`grep --hidden`** — include dotfiles/dot-directories in the recursive walk
   (off by default, bash no-dotglob semantics).
+- **`grep --max-count <N>`** — stop after N matching lines per file (GNU
+  semantics; `--max-count 0` matches nothing). Streaming single-file and piped
+  searches stop reading once the cap is hit, so it bounds work on large inputs.
 
 ### Fixed
 - **`sed -e <number>` is a loud error, not a silent drop.** A non-string `-e`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,17 @@ breaking entries are marked **BREAKING**.
 
 ## [Unreleased]
 
+### Added
+- **`grep --ftype <NAME>` / `--ftype-not <NAME>`** — filter the recursive (`-r`)
+  walk to (or away from) a named file type, e.g. `grep -r needle . --ftype rust`
+  (searches `*.rs`). Both are repeatable. `--ftype` is the kaish-wide file-type
+  filter (distinct from file/dir *entry-kind*). An unknown type name is a loud
+  exit-2 error, never a silent empty match.
+- **`grep --ftype-list`** — print the known file types and their globs as a
+  TYPE→globs table (works with `--json`); no pattern required.
+- **`grep --hidden`** — include dotfiles/dot-directories in the recursive walk
+  (off by default, bash no-dotglob semantics).
+
 ### Fixed
 - **`sed -e <number>` is a loud error, not a silent drop.** A non-string `-e`
   expression (e.g. `sed -e 5`) now exits 2 with a usage error instead of being
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,11 @@ breaking entries are marked **BREAKING**.
 - **`grep --max-count <N>`** — stop after N matching lines per file (GNU
   semantics; `--max-count 0` matches nothing). Streaming single-file and piped
   searches stop reading once the cap is hit, so it bounds work on large inputs.
+- **`glob --ftype <NAME>` / `--ftype-not <NAME>` / `--ftype-list`** — the same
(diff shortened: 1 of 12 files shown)


tobert:
## Review notes & follow-ups

### Review trail
- **Requested Gemini via kaibo** (read-only, drives a kaish shell over the repo). Provider was overloaded — two `503 UNAVAILABLE`, no auto-retry.
- **Fell back to DeepSeek** (our standing pick for kaibo reviews — it's caught regressions other models missed before). It read the four source files + traced the kernel arg-binder.

### What the review found
- **Design sound across all five scrutiny areas**: per-file `--max-count` in every path, the `hit_limit` vs `stopped_early` split, repeatable `--ftype` reading, loud unknown-type mapping, and the dirs-no-op interaction.
- **One real bug** → fixed in `00d5b4b`: when `--max-count` stops the chunked reader at a 256 KiB boundary that splits a UTF-8 multibyte char, `GrepLineScanner::finish()` validated the *artificially* truncated carry and raised a spurious exit-2 "binary data" error over a correct capped result — loud in the wrong direction. Guarded `finish()` on `hit_limit`; regression test verified **red-then-green**.
- **Dismissed — false positive**: "no tests." DeepSeek only saw the 4 attached source files, not the 18 kernel-routed tests in the two `*_search_features_tests.rs` files (which already cover its suggested list).
- **Judged benign, not acting**: `--ftype=` (empty value) → loud `unknown file type ''` exit 2. Correct loud behavior, not a bug.

### Follow-ups deferred to `docs/issues.md` (not blocking)
- **grep `--no-ignore`** — won't ship a flag that lies: the honest semantics of a per-call ignore-bypass under an embedder's `Enforced` scope (kaibo's preset) need a design decision first (does a user flag override the embedder's context-flood protection?).
- **Audit glob's *existing* `--no-ignore`** — it bypasses unconditionally regardless of scope; confirm that's intended vs. silently escaping `Enforced`. Feeds the same design.
- **`find --no-ignore`** — optional escape, only meaningful under `Enforced`; record-don't-build (keeps `find` POSIX).

### On ship
`docs/search-features-port.md` (the design + punch-list) is transient — delete it once this lands. The deferred items above already live in `issues.md`, so they survive that deletion.

PR #40, opened 2026-06-29
feat: execute_argv — argv-native kernel entry point

tobert:
## What & why

Adds `Kernel::execute_argv(name: &str, argv: &[Value]) -> Result<ExecResult>` — an **argv-native peer** of the string-native `execute(&str)`.

`execute(&str)` lexes and parses its input. A caller that already holds tokenized argv (a structured embedder like kaijutsu, a future busybox-style multicall binary) otherwise has to re-quote argv into a string just to have the lexer split it apart again — a round-trip that is **lossy** for typed values, since `to_argv()` stringifies `Value::Bytes`/`Value::Json`. `execute_argv` skips it.

This lands the load-bearing half of [`docs/multicall.md`](docs/multicall.md); the `kaish-multicall` binary and a `&[String]` convenience door are deferred (recorded in `docs/issues.md`).

## The key design decision

The design doc proposed a new `build_args_from_argv` that builds `ToolArgs` directly. I **rejected that** — a second binder is exactly the drift hazard the validation-builder unification fought. Instead the only new logic is a classifier `argv_to_args(&[Value]) -> Vec<Arg>` that maps each token to the AST `Arg` the lexer would produce for the equivalent minimally-quoted word. `execute_argv` then runs the resulting `Command` through the **existing** `execute_pipeline` → dispatch chain → `build_args_async` binder, verbatim.

- **One binder, no drift.** The argv door and the string door converge late, at the shared dispatch chain.
- **Typed passthrough.** `Expr::Literal(Value)` carries the value and evaluates by identity — it never becomes a `GlobPattern`/`VarRef`, so no glob/`$VAR`/`$()`/word-splitting can occur, and `Bytes`/`Json` land straight in `ToolArgs.positional`.
- **Tokens are literal** (single-quoted-word semantics). One shared-binder expansion still applies for consistency with the string door: a leading `~` expands against the session `HOME` (kaish expands `~` uniformly, even quoted).
- **Same tail.** Command resolution, `--json`, the confirmation latch, and `request_timeout`/watchdog all apply. The kernel *syntax* validator is skipped (argv has no shell syntax); per-tool `validate()`/clap still runs at dispatch.

## Cross-model review

Reviewed by **Gemini 2.5 Pro** (batch) and **DeepSeek** (interactive, via kaibo) — full transcripts posted as comments on PR #40. They converged tightly; here's the disposition:

**Fixed**
- **Classifier edge cases** (`-k=v`, `--=v`, `---`) that silently misbound into the binder → now literal positionals (house rule: loud/visible beats silent-wrong). The proptest re-pinned `-a:` → `ShortFlag("a:")`, so the rule landed as "short-flag body = alpha-led, no `=`".
- **`request_timeout` was ignored** (both rated BUG) → extracted the watchdog/timeout-race into a shared `run_under_watchdog(timeout, cancel, work)` helper that **both** doors now use; `execute_argv` honors `request_timeout` (124-on-deadline, cancel cascades SIGTERM/SIGKILL to children) for parity with `execute()`. All existing watchdog/cancellation suites stay green, confirming the extraction is behavior-preserving.
- **Tilde** — verified consistent with the string door, not a divergence; docs made honest; pinned by an equivalence test.
- **Test gaps** (malformed flags, WordAssign fallback) filled.

**Dismissed / deferred (with local verification)**
- Gemini's "skips RAII guards → `cd`/`export` leak global state" — **false positive**: those guards are conditional on per-call `ExecuteOptions` overrides (no-ops under defaults), so `execute_argv` matches `execute(default)`; `cd`/`export` persisting is correct shell behavior.
- "WordAssign ignores `past_double_dash`" — pre-existing **shared-binder** bug (hits the string door too) → recorded in `docs/issues.md` P4, not an `execute_argv` regression.
- Both confirmed the latch/`--json`/clap validation reach correctly and skipping the syntax validator is safe.

## Tests

- In-crate classifier suite incl. the **workspace's first proptest** (`classifier_matches_parser_on_clean_tokens`) — asserts `argv_to_args` agrees with the *real parser* on metacharacter-free tokens; stable across 40k cases; the three caught edges (`:A=0`, `00`, `-a:`) are pinned as regression seeds.
- Kernel-routed `execute_argv_tests.rs`: argv door ≡ string door over a corpus, literal-not-globbed, no `$VAR` interp, tilde consistency, typed `Int`/`Json` passthrough, WordAssign fallback, `--json`, exit-127, full latch round-trip, and `request_timeout` interrupting a hung command.

`cargo test --all` and `cargo clippy --all --all-targets` clean.

## Also in this PR (board hygiene)

Retires the **shipped** rg-features port entry from `docs/issues.md` and deletes the transient `search-features-port.md` (work landed in #38), narrating it in `docs/devlog.md` — folded in here to avoid two branches both editing `issues.md`/`devlog.md`.


tobert:
## 🤖 Review — Gemini 2.5 Pro (batch, deep-thinking)

Run against the full diff plus the reused-function context (`build_args_async`, `execute_command_depth`, `execute_pipeline`, the `Arg`/`Expr` enums) inlined, since batch jobs can't fetch files. Verbatim below. Disposition is in the PR description; in short: the classifier findings and the timeout finding were fixed; the "RAII guards" finding was a false positive (verified locally); the WordAssign-`--` finding is a pre-existing shared-binder bug recorded in `issues.md`.

---

### The Design Decision: Reusing `build_args_async`
The implementer's claim that reusing `build_args_async` avoids drift and safely converges the argv/string doors is **architecturally sound but currently flawed in execution**.

The fundamental mismatch is that `argv` elements are inherently "quoted" (literal strings, like `execve`), but `build_args_async` assumes `Expr::Literal` strings are eligible for shell expansions that don't require globbing. By forcing argv through the existing binder without a way to signal "this is a literal argv token," you inherit unwanted shell semantics. Furthermore, the `execute_argv` wrapper skips critical sandbox setup that the string door performs.

### [BUG] `execute_argv` skips watchdog and cancel token setup (Liveness/Safety)
* **Problem:** `execute_argv` calls `let _cancel = self.reset_cancel();` but completely skips setting up `exec_ctx.watchdog` and mirroring an embedder-provided cancel token.
* **Why it matters:** This is an agent shell. If an external command or builtin hangs, the lack of a watchdog means it will block forever, bypassing the shell's timeout mechanism. It also cannot be interrupted by the embedder.
* **Suggested fix:** `execute_argv` must accept `ExecuteOptions` (or use safe defaults) and replicate the watchdog and effective-cancel setup exactly as `execute_with_options_inner` does.

### [BUG] `execute_argv` skips RAII guards for vars/cwd/stdin (Hermeticity/Safety)
* **Problem:** `execute_with_options_inner` instantiates `vars/cwd/stdin/pipe-stdin` RAII guards before execution to isolate state. `execute_argv` skips these.
* **Why it matters:** Commands executed via `execute_argv` (e.g., `cd /tmp` or `export FOO=bar`) will permanently mutate the kernel's global state.
* **Suggested fix:** Wrap the `execute_pipeline` call in the same RAII guards used by the string door.

> _Maintainer note: false positive — those guards are conditional on per-call `ExecuteOptions` overrides and are no-ops under defaults, so `execute_argv` matches `execute(default)`; `cd`/`export` persisting is correct shell behavior._

### [BUG] Tilde expansion violates the "tokens are literal" argv contract
* **Problem:** `classify_argv_token` maps `"~/x"` to a literal positional; `build_args_async` unconditionally calls `apply_tilde_expansion` on it.
* **Why it matters:** If an embedder passes `["--message", "~foo"]`, it should remain `~foo`.
* **Suggested fix:** add a flag/variant so `build_args_async` skips `apply_tilde_expansion` for argv literals.

> _Maintainer note: verified this is consistent with the string door (kaish expands `~` even in quotes), so behavior kept and docs made honest rather than diverging the two doors._

### [RISK] `WordAssign` ignores `past_double_dash`
* **Problem:** `DoubleDash` sets `past_double_dash` which degrades `ShortFlag`/`LongFlag` to positionals, but the `WordAssign` arm does not check it. `["export", "--", "A=1"]` still parses `A=1` as a named `WordAssign`.
* **Suggested fix:** Update `build_args_async`'s `WordAssign` arm to check `past_double_dash`.

> _Maintainer note: pre-existing shared-binder bug (string door too) → recorded in `docs/issues.md` P4._

### [RISK] Empty keys in `Named` arguments (`--=v`)
* **Problem:** `--=v` becomes `Named { key: "", value: "v" }`.
* **Suggested fix:** Require non-empty key; fall back to `Positional`. _(Fixed.)_

### [RISK] Embedded `=` in `ShortFlag` (`-k=v`)
* **Problem:** `-k=v` maps to `ShortFlag("k=v")`; if `k` is a value flag, `bind_glued_short_value` may bind `=v`. _(Fixed — `-k=v` is now a literal positional.)_

### [TEST-GAP] Proptest `canonical()` collapse hides `WordAssign` fallback behavior
* Ensure an integration test covers `WordAssign` fallback for non-allowlisted commands (`["echo", "A=1"]`). _(Added.)_

### [NIT] `update_last_result` placement and Latch validation — **Correct.**
Placing `update_last_result` after `execute_pipeline` matches `execute_streaming_inner`. Skipping the kernel syntax validator is safe because the destructive-op confirmation latch and arg validation run at dispatch, which `execute_pipeline` reaches.

tobert:
## 🤖 Review — DeepSeek (interactive, via kaibo)

kaibo drove a read-only kaish shell over the actual files on disk (synth `deepseek-v4-pro`, explorer `deepseek-v4-flash`) and answered with `file:line` citations. Verbatim below (its line numbers are from kaibo's read and may be ±a few). Disposition is in the PR description.

---

The core idea — classify argv into `Arg` and reuse the existing `build_args_async` → dispatch chain — is **sound in principle**. The reuse of the single binder is the right call. However, the classifier had three concrete misclassification bugs, a tilde-expansion question, and a missing timeout surface.

### BUG: `-k=v` classifier misclassification → silent misbind
The classifier produced `Arg::ShortFlag("k=v")` for `-k=v`. The lexer tokenizes `-k=v` as `ShortFlag("k")` + `Eq` + `Ident("v")`, and the parser has **no `short_flag_with_value`** production, so `cmd -k=v` is a **parse error** in the string door. `build_args_async` would either bind `k = "=v"` (glued arm) or insert `"="` into `flags` (combined arm) — silent corruption. _(Fixed — now a literal positional.)_

### BUG: `---` classifier misclassification
`---` → `LongFlag("-")`. The lexer tokenizes `---` as `DoubleDash` + `MinusAlone` (`--` then a stdin-`-` positional). The classifier's single `LongFlag("-")` is wrong, and it also breaks the double-dash contract (no `Arg::DoubleDash` emitted). _(Fixed — long flags now require `--[a-zA-Z]…`; `---` is a literal positional.)_

### BUG: `--=v` classification → empty-key Named
`--=v` → `Named { key: "", value: "v" }`. An empty key is never valid. _(Fixed.)_

### RISK: Tilde expansion violates the "literal token" contract
`apply_tilde_expansion` is called on every positional/named value through `build_args_async`, so a literal argv `~/x` expands when `HOME` is set. `multicall.md` says tokens are "exactly like single-quoted words," but single-quoted words in bash are not tilde-expanded.

> _Maintainer note: verified `apply_tilde_expansion` runs post-eval and expands `~` even for single/double-quoted strings in the **string door** too — so `execute_argv` is consistent with `execute`, which is the equivalence contract. Kept the behavior, made the docs honest, pinned with a string-door-equivalence test._

### RISK: No timeout mechanism in `execute_argv`
`execute_argv` sets up no watchdog. Since `execute_with_options_inner` restores `ec.watchdog = None` after each call, `execute_argv` always sees `None` — a hung command cannot be interrupted by `request_timeout`. _(Fixed — shared `run_under_watchdog` helper; `execute_argv` honors `request_timeout`.)_

### Confirmed-safe: Cancel token flow
Traced: `Kernel::cancel()` → `self.cancel_token` → pipeline `ctx.cancel` → `ec.cancel` (synced by `dispatch_command`) → builtin ctx → external `wait_or_kill`. Works.

### Confirmed-safe: Latch, `--json`, per-tool clap parse
All run through the shared `execute_command_depth` path regardless of entry door. `rm.execute()` → `RmArgs::try_parse_from(args.to_argv())`; `GlobalFlags::apply_from_args` + `finalize_output` for `--json`; latch via `ctx.latch_result()`/`verify_nonce()`. The kernel-level pre-execution Validator is correctly skipped (it validates parsed AST; argv has none). The integration latch round-trip test is correct.

### Confirmed-safe: Concurrency and `$?`
`execute_argv` acquires `self.execute_lock` (same as `execute`); `update_last_result` placement matches the `Stmt::Command` arm.

### TEST-GAP: Missing classifier edge cases
`-k=v`, `---`, `--=v`, `-=` were untested; the proptest's `=`-containing tokens hit parser rejections and were silently skipped. _(Added explicit unit cases.)_

### NIT: Discarded `reset_cancel` return value
`let _cancel = self.reset_cancel();` — return value unused (the string path uses it). _(Resolved — now used as the watchdog's cancel token.)_

### Verdict
The design decision to reuse the single binder is **correct** — the `multicall.md` rationale holds, and the shared dispatch path is the right call. The bugs were all in the classifier, not the architecture.

PR #62, opened 2026-07-02
feat(collections): value/argv grammar seam (inert refactor)

tobert:
## What

Splits two narrow **value-position** seams out of the shared primary-expression parser, ahead of the follow-up PR that adds collection literals (`[a b c]`, `{k: v}`). This PR adds **no new syntax** — it is a provably-inert refactor that erects the wall so literals can appear only where a value is produced and never leak into argv or `for`-heads.

## Why this split

Collection literals must appear on assignment RHS and the `in`/`not in` collection operand, but **must not**:
- leak into argv (`ls [dog]` must stay a glob), or
- open up `for`-head items, which stay `$()`-only (bare `$VAR` splice is rejected by validator E012).

Making this a **structural** property (argv sites simply never call the value seam) beats a runtime context flag — a reviewer can see the invariant.

## The change

Two new pure-delegate combinators (both currently just call `primary_expr_parser()`):
- `value_expr_parser()` — assignment RHS: bash-style, `local`, env-prefix (previously `expr_parser()`).
- `value_primary_parser()` — the RHS (collection) operand of `in`/`not in`; the left needle stays on `primary_expr_parser`.

**Left untouched** (not literal positions): return/exit, file/string tests, comparison operands, here-strings, all argv, and `for`-head items via `expr_parser()`.

## Design provenance

Classification reconciled from two independent groundings (kaibo/deepseek + a planning pass) against the current tree. Part of the collections write-half milestone (`docs/arrays-and-hashes.md`); the next PR (PR-B) stops the lexer fusing `[`-leading glob runs and adds argv bracket reassembly + the `...` token.

## Verification

- `cargo test --all` — all green, **zero test edits** (the inertness proof)
- `cargo clippy --all --all-targets` — zero warnings

### DIFF
--- a/crates/kaish-kernel/src/parser.rs
+++ b/crates/kaish-kernel/src/parser.rs
@@ -1039,7 +1039,7 @@ where
         // prefixed-command form wins when a command follows.
         let env_prefix_assign = ident_parser()
             .then_ignore(just(Token::Eq))
-            .then(expr_parser())
+            .then(value_expr_parser())
             .map(|(name, value)| Assignment { name, value, local: false });
         let env_scoped = env_prefix_assign
             .repeated()
@@ -1116,7 +1116,7 @@ where
     let local_assignment = just(Token::Local)
         .ignore_then(ident_parser())
         .then_ignore(just(Token::Eq))
-        .then(expr_parser())
+        .then(value_expr_parser())
         .map(|(name, value)| Assignment {
             name,
             value,
@@ -1127,7 +1127,7 @@ where
     // The lexer produces IDENT EQ EXPR, so we parse it here
     let bash_assignment = ident_parser()
         .then_ignore(just(Token::Eq))
-        .then(expr_parser())
+        .then(value_expr_parser())
         .map(|(name, value)| Assignment {
             name,
             value,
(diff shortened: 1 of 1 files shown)


tobert:
**kaibo (deepseek) seam-placement audit — PASS.** Independent enumeration of every `primary_expr_parser`/`expr_parser` call site in the current `parser.rs` confirms the classification is complete and correct:

- All 5 rewired sites (assignment RHS ×3, `in`/`not in` RHS ×2) are genuine value positions.
- No value position missed; no non-value position wrongly rewired.
- `for`-head has **no transitive path** through either new combinator, so literals stay out of for-heads (validator E012 preserved).
- No `[`/`{` ambiguity: `[[ ]]` outer brackets are consumed before the inner value parser runs; case patterns / `${…}` are separate paths; bare `[` in argv/comparison stays a clean parse error.
- `return`/`exit` correctly left on `primary` (exit-code positions, not value output).

Deferred (not required by the design): `[[ $x == [a b c] ]]` (comparison RHS against a literal) stays a parse error — widen comparison RHS to the value seam only if a use case appears.

Inert refactor, `cargo test --all` green with zero test edits, clippy clean. Ready to merge.

PR #63, opened 2026-07-02
feat(collections): close the last two silent-stringify boundaries (D+E)

tobert:
Implements decisions **D** and **E** from `docs/arrays-and-hashes.md` — the last two places a live collection (`Value::Json(Array|Object)`) could silently stringify past a boundary via `value_to_string`, mirroring the shipped `structured_export_error` guard for OS-env export.

## Decision D — collection at the process edge is a loud error
A bare (un-interpolated) collection reaching an external command's argv or a redirect target now errors, pointing at `$(tojson $x)`. A quoted `"$x"` already reduces to `Value::String` via interpolation before the boundary, so `curl -d "$cfg"` is unaffected — only bare `curl -d $cfg` trips it.

Guarded at **every** external/redirect edge (a kaibo completeness audit drove the second commit — the first pass missed the passthrough builtins):
- `kernel.rs::build_args_flat` (production argv: Positional/Named/WordAssign) + `dispatch.rs::try_external` (test twin)
- `scheduler/pipeline.rs::eval_redirect_target` (all `>`/`>>`/`<`/`<<<` targets)
- **`spawn` / `exec` / `env` builtins** — each has its own argv/env stringify path that bypasses `build_args_flat`. The `env` case was a genuine silent-corruption path: `env /bin/true` with an exported record previously exited **0** with the collection JSON-serialized into the child env.

## Decision E — scalar test operators reject collection operands
In `[[ ]]`, a collection operand to `-z`/`-n`/`=~`/`!~`/`-eq`/`-ne`/`-gt`/`-lt`/`-ge`/`-le`/`<`/`>`/`<=`/`>=` is a loud Shape error (with a hint: `${#x}` for length, `-list`/`-record` for shape, `in` for membership). Applied via a single `guard_scalar_test_operands` call point in both the sync (`eval.rs::eval_test`) and async (`kernel.rs::eval_test_async`) evaluators. **`==`/`!=` are excluded** (they route to `values_equal`, which keeps structural collection-vs-collection equality and its own collection-vs-scalar loud error) and `in`/`not in` still accept a collection RHS.

## Verification
- Load-bearing red-toggle: disabling the env-export and spawn-nested guards reproduces exit-0-with-corruption / silent stringify; restored.
- `cargo test --all --features subprocess` (97 binaries) + `cargo test --all` + `cargo clippy --all --all-targets` (with and without `subprocess`) — all green.
- New tests through `kernel.execute()` (Linux-gated, `subprocess`): external argv, redirect target, spawn nested element, spawn record-as-argv, exec, env argv, env exported-collection.

Also files a P3 in `docs/issues.md` (separate bug class, not fixed here): `"$(cmd)"` quoted interpolation reads `.out`, so a command setting only `.data` (a collection) silently interpolates to `""` — a data-loss, not a stringify.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -134,6 +134,19 @@ breaking entries are marked **BREAKING**.
     default on a *subscripted* path — `${#u[tags]}`, `${cfg[port]:-8080}` — were
     briefly a loud "bind it first" error; they are now fully path-aware, see
     Added.)
+- **A bare collection reaching an external command's argv or a redirect
+  target silently JSON-serialized** (e.g. `curl -d $cfg` or `echo x > $cfg`);
+  both process-boundary sinks now refuse it with a `$(tojson $x)` hint, closing
+  the last two silent-stringify seams alongside the OS-env-export guard above.
+  A quoted `"$cfg"` is unaffected — string interpolation already renders
+  compact JSON before either sink sees it.
+- **Scalar-only `[[ ]]` test operators silently stringified a collection
+  operand** — `-z`/`-n`, `=~`/`!~`, ordering (`<`/`>`/`<=`/`>=`), and numeric
+  (`-eq`/`-ne`/`-gt`/`-lt`/`-ge`/`-le`) now refuse a list/record operand with a
+  loud Shape error naming the operator (e.g. an empty list no longer reads as
+  `-z`-true via its JSON text `"[]"`). `==`/`!=` already errored; `in`/`not in`
+  are unaffected (the one operator family that legitimately takes a
+  collection).
 - **`${x:-default}` no longer treats JSON `null` as present** — a variable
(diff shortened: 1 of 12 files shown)


tobert:
**kaibo (deepseek) completeness audit — drove the D-coverage.** Independent enumeration of every `value_to_string`/output boundary against the tree confirmed the primary sites (argv, redirect, `-z`/`-n`) and surfaced the passthrough-builtin gaps (`spawn`/`exec`/`env`) that a single-path guard would have missed. Display paths (echo, `"$var"` interpolation, builtin-clap bridge, heredoc, kaish-last) were explicitly verified as intentional compact-JSON rendering and left unguarded. `==`/`!=` structural-equality and `in`/`not in` collection-RHS paths confirmed preserved. All flagged gaps closed in the second commit.

PR #65, opened 2026-07-03
feat(#60): grep/sed/awk accept GNU BRE backslash-metas as an ERE superset

### LINKED ISSUE #60: grep: BRE `\|` alternation returns no matches (GNU-compat) — trips LLM agents hard
## Summary

kaish's `grep` does not treat `\|` as alternation in the default (BRE) mode — a pattern like `grep 'foo\|bar'` silently returns **zero matches** (exit 1) instead of matching `foo` OR `bar`. GNU grep supports `\|` as a BRE alternation extension, and it's deeply baked into how LLMs write grep. This is the single largest source of wasted tool calls when a model drives kaish (e.g. via kaibo's explorer).

## Repro (kaish-kernel 0.10.0, read-only mount over a real project)

```
$ grep -rn 'ConsultConfig\|ConsultOutput' src/      # BRE alternation
(no output, exit 1)

$ grep -rnE 'ConsultConfig|ConsultOutput' src/       # ERE
74 matches, exit 0
```

`-E` works correctly; only the BRE `\|` form silently no-matches.

## Why it matters (evidence)

Measured across kaibo explorer sweeps (deepseek driving `run_kaish`, tracked via OTLP `kaish.exit_code` spans): commercial models reflexively reach for `grep 'a\|b\|c'` — it's their default multi-term search. In one 41-call explorer sweep, **12 grep calls failed outright**, all `\|`, each followed by a single-term retry (thrash). The model does not reliably learn `-E` even when the preamble tells it to — the GNU habit is too strong. Fixing it in the tool is far higher-leverage than fighting it in every agent prompt.

Bonus: models also reflexively **batch** with `\|` across files — `grep 'a\|b' f1 f2 f3` — so supporting `\|` doesn't just remove failures, it turns those into the efficient single-call multi-term searches we otherwise can't coax out of them.

## Proposal

Support GNU-compatible `\|` alternation in BRE mode (and, for completeness, the other common GNU BRE backslash-metas if not already: `\+`, `\?`, `\{n,m\}`, `\(\)`). Matching GNU grep here removes a whole class of silent agent failures.

## Related (separate, lower priority)

`grep -r PATTERN FILE` with `-r` pointed at a **single file** (not a directory) also returns nothing, where GNU grep treats the file as a file. Same "silently empty, exit 1" failure shape. Mentioning for the same GNU-compat pass; can split to its own issue if preferred.

---
Found via kaibo's explorer-chattiness investigation (the new `kaish.exit_code` tool-span field made it visible). Happy to provide the full trace data.

tobert:
Closes #60.

## Why

Commercial LLMs reflexively write GNU BRE — `grep 'a\|b'`, `sed 's/\(a\)\(b\)/\2\1/'`, `awk '/cat\|dog/'` — and don't unlearn it from preamble text. Issue #60 measured `\|` as the single largest source of wasted agent tool calls (12 of 41 explorer greps in one sweep, each followed by a single-term retry). kaish had three different wrong answers: grep silently matched a literal `|` (zero results, exit 1), sed loud-rejected with the E006 teaching error, and awk silently no-matched.

## What

A shared rewriter, `regex_dialect.rs::bre_metas_to_ere`, turns the seven GNU BRE backslash-metas (`\| \+ \? \( \) \{ \}`) into their bare ERE operators before compilation, making kaish's regex a **superset**: both spellings mean the same operation. The rewrite is escape-aware (`\\` consumes as a unit, so `a\\|b` stays literal-backslash-then-alternation) and passes every other escape (`\.`, `\d`, `\b`, `\w`) through verbatim.

### Decisions

- **Default dialect is the superset; `-E` (grep) / `-E`/`-r` (sed) now mean *strict ERE*** — the escape hatch where `\|` matches a literal pipe. awk always rewrites (POSIX awk has no such flag and we didn't invent one). The narrow casualty: in the default dialect a backslashed meta is always the operator; match the literal character with a bracket class (`[|]`, `[)]`, `[{]`).
- **This reverses the sed ergonomics-pass decision** that erroring beats guessing: with the rewrite it's no longer a guess. `detect_bre_idiom` and its E006 error are deleted (**BREAKING** — scripts that relied on the error, or on `-E`-free `\|` matching a literal pipe, must use `-E`/`-r` or a bracket class).
- **awk field splitting stays gawk-compatible**: the rewrite runs *before* the single-char-FS-is-literal rule, mirroring gawk's demotion of `\|` to plain `|`. So `-F '\|'`, `FS="\\|"`, and `split(s, a, "\\|")` split on a literal pipe (NF=3 on `a|b|c`, matching gawk). The naive order — rewrite inside the multi-char regex branch — produced the empty-alternation regex `|` and silently split between every character (NF=7); caught in review, fixed with kernel-routed regression tests.
- **Compile errors carry a dialect hint** when the rewrite actually changed the pattern: `grep ':\)'` fails with `unopened group` *on a pattern the author never wrote* (`:)`), so the error appends "a backslashed |+?(){} is a GNU BRE operator in the default dialect — match the literal character with a bracket class like [)] or [|], or pass -E for strict ERE" (awk's variant omits `-E`).
- Bare ERE (`foo|bar`) and `-F` are untouched; `-F 'a\|b'` matches the verbatim 4-char text.

### Docs

CHANGELOG (Added + BREAKING Changed), `docs/LANGUAGE.md` sed section, `help` content (`awk.md`, `overview.md`), and the schema-visible `-E`/`-r` flag help (which previously claimed the flags were no-ops — `schema_from_clap` surfaces those doc comments to agents).

## Review

Self-review plus kaibo consult (deepseek, whole files attached, no diff). Both independently found the awk FS regression; kaibo additionally confirmed all 8 awk regex call sites route through the rewrite, sed validate/execute dialect parity, grep `-F`/`-E`/`-w` interaction, and complete E006 removal. Findings and follow-ups are posted as PR comments.

## Testing

- `regex_dialect.rs` rstest table for the rewriter (escape units, class interiors, trailing backslash).
- New kernel-routed binaries: `grep_bre_alternation_tests.rs` (11 tests: alternation, groups+quantifiers, intervals, `-E` strict, `-F` verbatim, dialect hint) and `sed_awk_bre_dialect_tests.rs` (11 tests: sed default/`-E`/`-r`, awk patterns/`~`/`gsub`, the three FS/split literal-pipe regressions, multi-char FS still ERE, dialect hints).
- sed unit tests rewritten from reject-assertions to translate-assertions; `jq_sed_validation_tests.rs` updated.
- `cargo test --all` and `cargo clippy --all --all-targets` clean; FS behavior verified against gawk 5.4.0 side by side.


tobert:
## Review commentary

Two-track review before this PR was opened: self-review (Claude, in-session, with live gawk 5.4.0 differential probes) and a kaibo consult (cast `deepseek`, whole files attached, no diff per our holistic-review convention). Both tracks **independently converged on the same headline bug**, which is fixed in this PR.

### Findings (all fixed in this PR)

1. **awk FS/`split()` regression** (both reviewers, independently). The single-char-FS-is-literal check ran on the *raw* separator, so the gawk literal-pipe idiom `FS="\\|"` fell into the multi-char regex branch, got rewritten to the empty-alternation regex `|`, and silently split between every character — NF=7 on `a|b|c` vs gawk's 3. Verified against gawk side by side (gawk warns "escape sequence \`\|' treated as plain \`|'" then applies the POSIX single-char rule). Fix: rewrite *before* the single-char check, in both `split_record` and `split()`. Silent-wrong field splitting is exactly the failure class this PR exists to kill, so this one mattered.
2. **Stale schema-visible flag help** (kaibo). grep's `-E` doc comment still said "No-op: Rust's regex crate is always extended"; sed's said "compatibility no-op… BRE groups are rejected with a hint." Both false after this change — and `schema_from_clap` surfaces those comments as agent-facing parameter help, so this was wrong documentation shown to every agent, not a comment nit.
3. **Test gaps** (kaibo): no `sed -r` coverage (the short alias binds through a different kernel arg path than `-E`), no awk FS/`split()`/`gsub()` BRE coverage — any FS test would have caught finding 1 — and no `-F`-with-BRE-spelling pin. All added.
4. **Confusing engine errors on rewritten patterns** (self-review). `grep ':\)'` failed with `unopened group` pointing at `:)` — a pattern the author never wrote. Fixed with a dialect hint appended only when the rewrite actually changed the pattern; awk's variant omits the `-E` advice it doesn't have (pinned by test).

### What was checked and found sound

- The rewriter: `\\` consumed as a unit (`a\\|b` stays literal-backslash-then-alternation), non-meta escapes verbatim, trailing backslash passthrough, bracket-class interiors equivalent under rewrite (`[\|]` ≡ `[|]`).
- All 8 awk regex call sites route through the rewrite (6 via `compile_ere`, FS + `split()` directly); no missed `Regex::new`.
- sed validate/execute dialect parity; `extended` threaded through addresses and `s///` only (`y///`, `a/i/c`, `d/p/q` don't touch regex).
- grep validate/execute parity; `-F` → `regex::escape` before `-w` wraps `\b…\b`; `-F` bypasses the rewrite entirely.
- `detect_bre_idiom`/E006: zero remaining references.
- GNU `\<`/`\>` word boundaries already work (Rust regex supports them natively) — no gap there.
- Formerly-literal escapes that now fail (`:\)`, `a\{`) fail **loudly** (E005 + hint), never silently.

Gates at HEAD: `cargo test --all` and `cargo clippy --all --all-targets` clean.

tobert:
## Follow-ups (not blocking this PR)

Small residuals noted during review, left out deliberately to keep the change focused:

1. **awk FS/`split()` compile errors don't carry the dialect hint.** The `invalid FS regex`/`invalid split() separator regex` paths report the engine error only. Reaching them via a rewritten meta requires a multi-char separator that's *still* invalid after the rewrite — rare, and always loud — but the hint would be consistent. One `append_dialect_hint(…, None)` call each if we care.
2. **Those same errors display the rewritten separator, not what the user typed** (e.g. `invalid FS regex "a("` for input `a\(`). Cosmetic; the hint from (1) would explain the delta if both land together.
3. **`sed -i` still conflicts with GNU's glued `-i.bak`** (pre-existing, documented in LANGUAGE.md) — unrelated to the dialect work but adjacent in the sed surface; noting so it isn't assumed covered by this pass.
4. **Consider a `help regex` fragment.** The dialect story (superset default, strict `-E`, bracket-class literals) now lives in `help sed`, `help awk`, grep's schema, and LANGUAGE.md; a single composable fragment would keep them from drifting. Fits the `kaish-help` concept taxonomy.

Items 1–2 are one small PR if wanted; 4 is a docs-shape question for the next help pass.

PR #68, opened 2026-07-03
fix: standalone [[ ]] writes $? + awk separator errors name the raw form

tobert:
Two follow-ups worked while the collections test-coverage batches ran.

## Standalone `[[ ]]` never wrote `$?` (P1)

`[[ 1 = 2 ]]; echo $?` printed `0` — the previous command's status. `Stmt::Test` evaluated the test and returned the right `ExecResult`, but never called `update_last_result`, so any following `$?` read was stale. Agents write `[[ -f x ]]; ok=$?` constantly; that pattern silently captured the wrong status. Found during PR-D testing (recorded in the 2026-07-02 signoff), reproduced on current main before fixing.

The fix mirrors `Stmt::Command`: write the result to `$?` and honor suppressible errexit. The composition is safe by construction — `&&`/`||` chain arms already suppress errexit around their left operand, and `if`/`while` conditions evaluate as *expressions*, never through the statement arm — so `[[ … ]] && cmd`, `set -e` scripts, and loop conditions all behave like bash. Pinned with five `shell_compat!` tests, all verified against real bash (`KAISH_BASH_COMPAT=1`): failure writes 1, success writes 0, capture into a var, `&&` gating, and the `set -e` trip on a bare failing test.

## awk separator errors name the raw form (PR #65 follow-ups 1+2)

Invalid FS/`split()` separator errors reported the internally ERE-rewritten separator (e.g. `xx(` for input `xx\(`) with no explanation. They now name the separator as the user wrote it and append the GNU-BRE dialect hint when the rewrite changed the pattern (awk's variant offers only the bracket-class spelling — no `-E` to suggest). Pinned for both the FS and `split()` paths.

## Testing

`cargo test -p kaish-kernel` (4k+ tests) and `cargo clippy -p kaish-kernel --all-targets` clean. Workspace-wide gates couldn't run at commit time — the working tree carries a parallel session's in-progress kaish-help edits (uncommitted `SyntaxSection` topic work), which fail to compile independently of this change; nothing in this PR touches kaish-help.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -163,6 +163,14 @@ breaking entries are marked **BREAKING**.
   demotion of `\|` to plain `|`).
 
 ### Fixed
+- **A standalone `[[ ]]` test now writes `$?`** — `[[ 1 = 2 ]]; echo $?` printed
+  `0` (the previous command's status; the test never stored its result), so
+  `[[ -f x ]]; ok=$?` patterns silently read a stale status. A bare test
+  statement now sets `$?` and honors `set -e` like any command, matching bash;
+  `&&`/`||` operands and `if`/`while` conditions are unaffected.
+- **awk invalid FS/`split()` separator errors name the separator as written**
+  (not the internal ERE-rewritten form) and carry the GNU-BRE dialect hint when
+  the rewrite is the likely culprit.
 - **Collection read-access silent-corruption traps** (found by the collections
   milestone review) — five cases that returned a plausible wrong answer instead of
   erroring, all now loud:
--- a/crates/kaish-kernel/src/kernel.rs
+++ b/crates/kaish-kernel/src/kernel.rs
@@ -2520,11 +2520,24 @@ impl Kernel {
             }
             Stmt::Test(test_expr) => {
                 let is_true = self.eval_test_async(test_expr).await?;
-                if is_true {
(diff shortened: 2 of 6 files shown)


tobert:

Both changes verified sound across the composition surfaces:

- **`&&`/`||` operands**: chain arms suppress errexit around the left side before recursing, so a failing `[[ ]]` operand returns normal flow; the double `$?` write (Test arm + chain arm) is the same value under the kernel's execute_lock — benign. A failing test as the *right* operand under `set -e` propagates the exit, matching bash.
- **`if`/`while` conditions**: evaluate via `eval_expr_async`, never reach the statement arm.
- **`$()` bodies**: scope is snapshot-restored around the block and `$?` is set from the block result afterward, so the inner write is a harmless no-op.
- **Functions**: a failing bare `[[ ]]` under `set -e` now exits the function body early — a behavior change that *matches bash* (previously execution continued past the failed test). More correct, not a regression.
- **Background/forked workers**: fully isolated scopes (fork clones the whole scope), no race on the `$?` write.
- **awk raw-separator threading**: capture-before-rewrite is correct in both `split_record` and `split()`; no path reports the rewritten string; the hint fires only when the rewrite actually changed the separator.

One reviewer wobble, discounted: it claimed scatter doesn't exist in the codebase (it does) — but the isolation argument holds regardless since scatter workers run on forked kernels with cloned scopes.

PR #70, opened 2026-07-03
test(collections): value-model invariant suite from the cross-model review

tobert:
Test-only PR: 30 new kernel-routed tests in `collections_value_model_tests.rs` plus one Linux-gated external-boundary test in `external_command_tests.rs`. No production code changes.

## Provenance

After #66/#67 landed the write half, we fanned out a coverage question to two independent reviewers: a **gemini-pro batch** (13 whole files attached — design doc, lexer, parser, eval/scope, the five collection builtins, all three test files) asked for the most important few tests for design confidence, and a **fable deliberation** (kaibo dossier → offline max-thinking) hunted edge cases adversarially. Both independently converged on the same hole: **the load-bearing value model had zero integration tests** — copy-on-assign, function-scope × mutation, RMW ordering, iteration snapshots, and the round-trip law the design doc declares "test-pinned" (it wasn't, until now).

## What's pinned

- **Copy-on-assign aliasing, both directions** — mutate original, mutate copy, extracted nested sub-values, `push` on a copy, key-insert on a copy. The named threat: a future "make clones cheap" Rc/Arc refactor would corrupt both bindings while the rest of the suite stayed green. These are the tripwire.
- **Function scope × mutation** — documented write-through (`f() { xs[0]=99; }` reaches the caller), the `local xs=$xs` protection idiom, `push` through a frame. Bonus finding: fable's scariest edge (bare `local ys`, then a subscript write silently punching through to the caller) turns out to be **structurally impossible — `local` without an initializer doesn't parse**. Pinned as a guard so any future bare-`local` support forces that interaction to be decided deliberately.
- **Read-modify-write** — `xs[0]=${xs[-1]}`, record RHS, `push ys ${ys[-1]}` appending exactly once, swap-via-temp. RHS evaluates against pre-write state.
- **Iteration snapshot** — growing or replacing the source inside `for v in $(values $xs)` changes neither the trip count nor the yielded elements.
- **Round-trip law** — `fromjson "$(tojson $x)"` through the `$()` and interpolation boundaries: nested structure, empty `[]`/`{}` keep their types, key insertion order, integer fidelity past 2^53, unicode/empty/space keys.
- **Decision B + key typing** — `$((xs[i]))`, `${xs[$i]}`, and `$((xs[$i]))` agree from one script (the two resolvers live in different modules and can drift); `"1"` and `"01"` are distinct record keys (no numeric normalization); bareword digits are integer indices even on records (loud, with the quoted-key fix-it); quoted numerics are string keys even on lists (loud). Stricter than the review panel's gawk-shaped assumptions — kaish refuses instead of coercing, both ways.
- **Boundaries** — `echo ...$xs` is a parse error; `[[ x in * ]]` is the loud scalar-RHS error (no glob expansion); `"$user.name"` leaves `.name` literal; a shape error inside a string interpolation stays loud (never renders partial output); `X=[1 2] /bin/true` refuses at the Decision-D boundary with the `tojson` hint (the guard fires at spawn, so the test needs a real binary — Linux-gated, `subprocess`-feature file).

## Method

Every assertion was probed against the live binary before being written — all behaviors were already correct, so this suite is regression tripwires, not fixes. Behaviors the review flagged as **unratified** (RHS-snapshot semantics, iteration materialization timing, numeric-key non-normalization, scatter/gather element typing) are deliberately not asserted; they're queued for design-doc ratification first.

## Testing

`cargo test --all` and `cargo clippy --all --all-targets` clean; `--features subprocess` external suite 36/36.

### DIFF
--- a/crates/kaish-kernel/tests/collections_value_model_tests.rs
+++ b/crates/kaish-kernel/tests/collections_value_model_tests.rs
@@ -1,0 +1,464 @@
+//! Value-model invariants for native collections (lists & records).
+//!
+//! Provenance: the 2026-07-03 cross-model coverage review (gemini-pro batch +
+//! fable deliberation over a kaibo dossier) — both independently flagged that
+//! the *load-bearing* semantics here had zero integration tests: copy-on-assign
+//! aliasing, function-scope × mutation, read-modify-write ordering, iteration
+//! snapshots, and the fromjson/tojson round-trip law the design doc declares
+//! "test-pinned". Every test pins RATIFIED design (docs/arrays-and-hashes.md);
+//! behaviors the review flagged as unratified are deliberately NOT asserted
+//! here — they're queued for Decisions-section ratification first.
+//!
+//! The shared failure mode these guard against is the project's cardinal sin:
+//! a plausible-but-wrong value with no error. In particular, a future
+//! "make clones cheap" Rc/Arc refactor of the value model would corrupt both
+//! bindings in every aliasing test below while the rest of the suite stays
+//! green — that's the tripwire.
+//!
+//! Kernel-routed via `KernelConfig::isolated()` (pure data, no localfs), same
+//! harness as `collection_access_tests.rs`.
+
+// Test-fixture code: unwrap/expect on known-good setup is the idiom here.
+#![allow(clippy::unwrap_used, clippy::expect_used)]
+
+use std::sync::Arc;
+
+use kaish_kernel::{Kernel, KernelConfig};
+
+async fn setup() -> Arc<Kernel> {
+    Kernel::new(KernelConfig::isolated().with_skip_validation(true))
+        .expect("kernel")
+        .into()
+}
+
+/// Run a script, returning (trimmed stdout, exit code, err text).
+async fn run(k: &Kernel, script: &str) -> (String, i64, String) {
+    let r = k.execute(script).await.expect("kernel execute");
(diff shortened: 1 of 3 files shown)


tobert:

Reviewed against the design doc, the three sibling test files, and the harness config. Verdict: 29 of 30 tests pin ratified design; findings below, all addressed in the follow-up commit.

- **Vacuous-pass risk (fixed):** the bare-`local` guard and the argv-spread rejection asserted only "some error occurred" — a future change that accepts the syntax and fails later at runtime would have passed both silently. Both now assert a parse-level rejection specifically.
- **One deliberate implementation pin (now declared):** `local_without_initializer_does_not_parse` guards current implementation behavior, not ratified design — the write-through safety property depends on it, so if bare `local` is ever accepted, the test forces a deliberate decision. The file header now names it as the explicit exception to the "ratified only" rule.
- **Attribution fix:** the write-through comment quoted "regardless of `local`" as if from the design doc; that phrasing came from the review dossier. The doc actually ratifies shared-scope functions + copy-on-assign, from which write-through follows — comment corrected.
- **Redundancy check:** no duplicates with the three sibling files; three tests partially overlap `collection_access_tests` but tighten error-message assertions or cover different input shapes (quoted `"1"` vs bareword, glob-shaped `*` vs plain string RHS).
- **Harness check:** `with_skip_validation(true)` weakens nothing here — all pins are runtime semantics or parser-level; the lvalue sibling file (which pins validator codes E016/E017) correctly keeps validation on instead.

PR #72, opened 2026-07-03
docs(collections): ratify the three review-flagged semantics; flag scatter typing

tobert:
Docs-only. The 2026-07-03 cross-model coverage review (gemini-pro batch + fable deliberation) found four collection behaviors that were true by code structure but never ratified in `arrays-and-hashes.md`. PR #70 pinned three of them with tests; this PR gives them Decisions entries so the tests defend ratified design rather than accident:

- **RHS evaluates against the pre-write state** — `xs[0]=${xs[-1]}` reads the old tail; `push ys ${ys[-1]}` appends exactly once. Load-bearing the moment anyone optimizes the clone-and-swap into in-place mutation.
- **Iteration materializes at loop entry** — mutations inside `for v in $(values $xs)` never feed back into trip count or yielded elements; a lazy/streaming refactor would break the contract.
- **Record keys are exact strings** — `"1"` ≠ `"01"`, never integer-parsed; strict no-coercion subscripts in both directions with the fix-it errors.

The fourth — **scatter/gather element typing** — is genuinely open (workers receive stringified elements today, so `1`/`"1"` conflate silently: the one currently-silent wrongness left from the review). Recorded as an Open decision with both options (typed workers matching the `$(values …)` precedent, vs. a ratified loud text boundary) and filed as issues.md P2 for the decide-then-implement pass.

### DIFF
--- a/docs/arrays-and-hashes.md
+++ b/docs/arrays-and-hashes.md
@@ -505,9 +505,46 @@ for the record.
   does not exist there. The lexer has no `..`/`...` handling at all — `...` is a new token,
   meaningful only inside `[ ]` value context.)*
 
+- **Write ordering — the RHS evaluates against the pre-write state.** *(Ratified 2026-07-03,
+  from the cross-model coverage review.)* In `xs[0]=${xs[-1]}` the RHS is fully evaluated
+  before `walk_write` clones-and-swaps the root, so a self-referential write reads the old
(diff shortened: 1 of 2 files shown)


tobert:
Superseded by the rebased #76 — the force-push hook made updating this branch's history more trouble than a fresh branch.

PR #74, opened 2026-07-03
fix(collections): loud subscripted-length on unset roots; record keys interpolate

tobert:
The two verified bugs from the 2026-07-03 cross-model coverage review (both reproduced against the live binary before fixing, both silent-wrong):

## `${#nope[items]}` silently returned 0

`resolve_length`'s bash-parity arm (`${#unset}` → 0) didn't distinguish bare roots from subscripted paths, so a typo'd variable name in `while [[ ${#queue[items]} -gt 0 ]]` looped zero times forever with no diagnostic — while the analogous bare access `${nope[items]}` errored loudly. An unset root under a **subscripted** path is now a loud `undefined variable` naming the path; bare `${#unset}` stays 0, pinned by a separate test so the two forms can't drift together. (fable finding #1; the review flagged it as "current bug or unratified — decide, then pin"; decided loud.)

## `{"$k": 8080}` silently created a literal `"$k"` key

The record-literal parser took `Token::String` raw, so double-quoted keys were the one place in the shell where `"$k"` didn't interpolate. They now resolve like every other double-quoted string: a new `RecordKey::Interpolated(Vec<StringPart>)` rides the existing StringPart machinery in both eval sites (sync `eval.rs` + async `kernel.rs`), and a pure-literal parse folds back to `Quoted` so plain `{"content-type": x}` keys carry no eval cost. `${k}_num` composites work; **single quotes remain the literal-`$` escape hatch** (`{'$k': v}` keys `$k`, unchanged); an unset var in a key expands to `""` per the ratified string-interpolation rule — pinned with a test so the choice is documented. (gemini finding #10, which flagged the unratified behavior; decided interpolate-for-consistency.)

## Docs

`LANGUAGE.md` and the collections help fragment gained the key-quoting line (`syntax.md` regenerated via `regen_syntax`); CHANGELOG `Fixed` bullets for both.

## Testing

Two new length tests in `collection_access_tests.rs`, four new key tests in `collection_literals_tests.rs` (interpolated, braced composite, single-quote literal, unset-var-empty). Full `cargo test --all` + `cargo clippy --all --all-targets` clean in the worktree; the one background-run flake (`external_command_tests`, load blip during concurrent worktree builds) re-ran green 5/5.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -172,6 +172,14 @@ breaking entries are marked **BREAKING**.
   demotion of `\|` to plain `|`).
 
 ### Fixed
+- **`${#path}` on an undefined subscripted root is now loud** — `${#nope[items]}`
+  silently returned `0` (the bash-parity `${#unset}` → 0 rule leaked into
+  subscripted paths), so a typo'd name in a length-guarded loop spun zero times
+  with no diagnostic. Bare `${#unset}` stays `0`.
+- **Double-quoted record-literal keys now interpolate** — `{"$k": v}` resolves
+  `$k` like any double-quoted string instead of silently creating a literal
+  `"$k"` key. Single quotes keep a literal `$` (`{'$k': v}`), same quoting
+  contract as everywhere else.
 - **A standalone `[[ ]]` test now writes `$?`** — `[[ 1 = 2 ]]; echo $?` printed
   `0` (the previous command's status; the test never stored its result), so
   `[[ -f x ]]; ok=$?` patterns silently read a stale status. A bare test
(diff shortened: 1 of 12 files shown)


tobert:

Both fixes verified sound; full pass over path shapes, callers, and consumers:

- **`segments.len() <= 1` guard**: right discriminator for every `VarPath` shape — bare root → 0 (bash parity), subscripted/dynamic → loud, degenerate empty → 0. The dynamic-subscript case (`${#xs[$undef]}` where the *subscript* var is unset) also lands loud, which is the correct tightening: a length you can't resolve reported as 0 would be a lie.
- **All `resolve_length` callers checked**: the three real callers propagate the new error; the two reduced-sync pipeline callers already swallow errors (the documented P3 gap) — no regression, and the silent-0 they used to produce for this case is gone.
- **Bonus bug fix found in review**: the old raw-token path leaked the lexer's `__KAISH_ESCAPED_DOLLAR__` marker into keys — `{"a\$b": v}` literally created a key containing the marker string. The new `parse_interpolated_string` route resolves it to `a$b`. A previously-broken spelling now works; no previously-working key changes behavior.
- **Sync and async eval sites agree** on error propagation for `$()` inside a key; all four `RecordKey` consumers (parser, both evals, sexpr) handle the new variant.

Verdict: LGTM.

PR #75, opened 2026-07-03
feat(help): `help regex` — one-screen dialect reference (PR #65 follow-up)

tobert:
The last open follow-up from PR #65: the regex-dialect story (ERE everywhere, GNU BRE superset, escape hatches) lived in four places — grep schema, sed schema, awk help, LANGUAGE.md — with no single surface an agent could be pointed at.

New `regex` syntax_section in `kaish-help`, riding #69's `render_syntax_section` mechanism, so `help regex` works simply by naming the section (it also appears in `help syntax` and the generated `syntax.md`).

Per the design direction: **token-efficient, and written to wake the model's ERE weights** — the fragment leads with *working ERE idioms* (alternation, capture groups, quantifiers, anchored alternation in awk) before mentioning BRE at all, so a model reading it gets primed toward the native dialect rather than the compat layer. Then one line each for: the accepted GNU BRE spellings, bracket-class literals, `-E`/`-r` strict mode, `-F`, extra escapes (`\d \w \b \< \>`), and the no-backreferences-in-patterns rule. One screen total.

Every example line in the fragment was executed against the real binary and produces the documented behavior.

`syntax.md` regenerated (drift test green); CHANGELOG `Added` bullet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,10 @@ breaking entries are marked **BREAKING**.
   test line (a prior regression against the shell's own teaching-note rule).
 
 ### Added
+- **`help regex`** — a compact syntax section (also in `help syntax`/`syntax.md`)
+  covering the regex dialect in one screen: ERE everywhere (grep/sed/awk), the
+  accepted GNU BRE spellings, the bracket-class/`-E` escape hatches for literal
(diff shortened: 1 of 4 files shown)


tobert:
Superseded by the rebased #77 (same content, rebased over #74's devlog/CHANGELOG lines).

Issue #78, opened 2026-07-03
xargs builtin — POSIX-flavored parallel/batch execution


tobert:
Counterpart to the scatter/gather JSONL redesign (#73): scatter/gather is becoming the kaish-native *typed* parallel map (JSONL result records, typed `ITEM`), which deliberately drops any pretense of POSIX text ergonomics. `xargs` is the planned home for the POSIX flavor: bare lines in, argv out, familiar flags (`-n`, `-P`, `-I{}`, `-0` subset per the 80% rule), bare-stdout results.

Scope sketch (not designed yet):
- newline-delimited stdin (and `-0`), items appended to the command argv
- `-P N` parallelism sharing the scatter runner's fork machinery
- exit codes per the xargs contract (123 on any failure) — which is also the prior the scatter/gather panel is testing
- no typed `.data` involvement; this tool is the text boundary

Design pass + model panel before implementation, same as #73.

PR #79, opened 2026-07-03
feat(#73): scatter/gather goes typed-JSONL — the kaish-native parallel map

### LINKED ISSUE #73: scatter/gather element typing — decide, then implement
`scatter` stringifies typed elements: a list of records reaches workers as compact-JSON text (`${ITEM[id]}` errors loudly), and `1`/`"1"`, `true`/`"true"`, `null`/`"null"` conflate silently inside workers — the one currently-silent wrongness left from the 2026-07-03 collections coverage review (fable finding #6).

The decision is recorded as an Open decision in `docs/arrays-and-hashes.md`: (a) workers receive typed values, matching the `for v in $(values …)` precedent (review's preference), or (b) scatter is ratified as a text boundary like external argv, refusing typed elements loudly. Gather's output typing rides the same call.

Once decided: implement + pin with kernel-routed tests, including `typeof $ITEM` and the conflation cases.

tobert:
Closes #73. **BREAKING** (pre-1.0), designed and ratified on the issue, panel-validated before implementation per `docs/designing-syntax-with-llms.md` (5-model task battery + 2 hostile frontier reviews; [raw data, secret gist](https://gist.github.com/tobert/ae4caa0075dab159934067735d7127e5)).

## What

scatter/gather becomes the kaish-native **typed parallel map**; a future `xargs` (#78) owns the POSIX bare-lines flavor.

**Ingress — typed `ITEM`.** Items stay JSON values through the fan-out; workers bind via `json_to_value_no_envelope` — the for-loop's exact conversion — so `${ITEM[id]}` subscripts records, numbers stay numbers, and the `1`/`"1"` conflation (the original #73 bug) is dead. Guards: a single non-array object is a loud error with a select-the-array hint (`did you mean jq '.jobs'?`), `null` items and binary input are loud, blank text lines are skipped, empty input exits 0.

**Egress — JSONL result records.** One compact record per worker, item order, **every worker including failures** — which deletes the old line mode's can't-represent-a-failure short-circuit hack:

```jsonl
{"i":0,"item":{"id":1,"host":"web1"},"ok":true,"code":0,"out":"restarted","err":""}
{"i":1,"item":{"id":2,"host":"db1"},"ok":false,"code":7,"out":"","err":"connection refused"}
```

`i`/`item`(typed)/`ok`/`code`/`out`(raw, trailing newline stripped)/`err`(always present — omit-empty on the most-read field would make `${r[err]}` a loud missing-key on success rows) on every row; `data` (worker's structured output, typed) and `timed_out:true` when present; timeout rows report `code` 124.

**One source, three views:** JSONL text (default) · `gather --json` one array (both tools now declare `owns_output`, so the kernel-wide flag reaches the tool — `--format` deleted) · typed `.data` (so `for r in $(… | gather)` iterates records and post-gather stages get typed stdin).

**Exit codes A′:** `0` all ok · `123` any worker failed, partial or total (rows carry which) · `2` usage. The panel split by tier here — cheap models guessed 0, frontier guessed 123 — and fable caught that the original proposal A *fought the prior it borrowed* (real xargs 123 covers all-failed too), which produced A′.

**Removed:** `--format` (folded into `--json`), `--first` (both hostile reviews independently: it's `head` on JSONL unless it cancels workers; the race-to-N-successes idea is a filed follow-up). **Added:** `--lines` — raw successful outputs in item order, hard error with no partial text if any worker failed (the old safety property as an explicit flag).

## The one post-panel discovery

kaish `jq` is **array-as-document** (`seq 1 3 | jq add` depends on it), so the gather idiom is `jq -r '.[] | select(.ok) | .out'` — *not* the real-jq streaming form all five panel models wrote one-shot. The docs teach `.[]` prominently in `help scatter`; if agents thrash on it in practice, a jq streaming affordance is the revisit point (noted in devlog).

## Testing

New `scatter_gather_jsonl_tests.rs` (16 kernel-routed tests): row schema/order, failure rows + 123, typed record subscripting, conflation kill, worker `data`, `--json` array, typed for-iteration, jq `.[]` streaming, post-gather typed stdin, `--lines` happy/hard-error, object/null/blank/empty ingress guards, timeout rows (124 + gather 123). Updated pins: `scatter_single_json_tests` (object→loud, deliberate reversal of the P2-sweep pin), `cancellation_tests` (timeout row shape), `json_sweep_tests` (`--json` instead of `--format json`). Every doc example verified against the binary. `cargo test --all` + `clippy --all-targets` clean; subprocess-gated suites 16/16 + 36/36.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,23 @@ breaking entries are marked **BREAKING**.
 ## [Unreleased]
 
 ### Changed
+- **BREAKING: scatter/gather is now the typed parallel map** (GH #73,
+  panel-validated). `gather` emits one JSONL result record per worker, in item
+  order, failures included — `{"i":N,"item":<typed>,"ok":…,"code":…,"out":…,
+  "err":…}` plus `data`/`timed_out` when present (timeout → `code` 124). The
+  same records are the structured data (`for r in $(… | gather)` iterates typed
+  records; kaish `jq` receives them as one array — stream rows with `.[]`), and
+  `gather --json` renders one JSON array. Worker bindings are TYPED: a JSON
+  array fans out element-by-element with real types (`${ITEM[id]}` subscripts a
+  record; `1` and `"1"` no longer conflate). Exit codes: `0` all ok · `123` any
+  worker failed · `2` usage. `gather --format` and `--first` are REMOVED;
+  new `gather --lines` emits raw successful outputs and hard-errors if any
+  worker failed. Scatter ingress errors loudly on a single non-array object
+  (with a select-the-array hint), `null` items, and binary input; blank text
+  lines are skipped. A future `xargs` builtin (GH #78) will carry the POSIX
+  bare-lines flavor.
+
+### Changed
 - **BREAKING:** **`sed` no longer rejects BRE idioms with a hint** — `\|`,
   `\(…\)`, and `\{N,M\}` used to be a loud `E006` error; they now behave as
   alternation/groups/intervals (see Added). Scripts that relied on the error, or
--- a/crates/kaish-help/content/en/scatter.md
+++ b/crates/kaish-help/content/en/scatter.md
@@ -1,34 +1,101 @@
-# 散/集 (Scatter/Gather) — Parallel Processing
+# 散/集 (Scatter/Gather) — Parallel Map over Structured Data
 
 ## Syntax
 
 ```
-input | scatter [--as VAR] [--limit N] [--timeout DUR] | command | gather [--first N] [--format lines|json]
+input | scatter [--as VAR] [--limit N] [--timeout DUR] | worker… | gather [--lines|--json]
 ```
+
+## The model
+
+scatter fans structured input out to parallel workers; gather collects one
+**JSONL result record per worker**, in item order, **every worker — failures
+included**:
+
+```jsonl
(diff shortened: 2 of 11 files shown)


tobert:

Full pass over the six risk areas; **no blocking issues**:

- **post_gather short-circuit**: correct — on failure the gathered result (rows + err summary) returns directly; single caller, nothing depends on the old `from_output` shape.
- **Old-contract sweep**: zero remaining references to `item: String`, `GatherOptions.first/format`, or the `Vec<String>` extract_items across all crates.
- **Worker scope aliasing**: safe — Arc-COW frames + sequential clone-in-loop means each worker's first `set()` deep-copies; no shared mutable state between workers.
- **Parent-cancel vs timeout rows**: distinguishable (timeout → `code:124` + `timed_out:true`; parent cancel → the interrupted command's code, no flag). The residual ambiguity (genuine code-1 failure vs cancellation) is inherent to external cancel, same as POSIX `kill`.
- **Standalone `gather --json`**: flag is silently a no-op in the marker's passthrough mode (`owns_output` blocks kernel rendering; the real path is the pipeline runner). Mildly surprising, low-risk — noting here rather than adding standalone-mode ceremony; the tool doc already says the runner owns the real behavior.
- **Help content**: every example in `help scatter` verified against the implementation — including the kaish-jq `.[]` streaming idiom, the typed for-loop row iteration, and the exit-code table. No drift.

PR #81, opened 2026-07-03
fix(vfs): mount /dev unconditionally in Kernel::with_backend

tobert:
## Summary

- Custom-backend embedders (kaijutsu's shape) built their kernel via
  `Kernel::with_backend`, which never ran the `setup_vfs()` path that mounts
  `DevFs` at `/dev` for `Kernel::new`/`Kernel::transient` — so `/dev/null`,
  `/dev/zero`, `/dev/random`, `/dev/urandom` were simply absent. Worse: in
  kaijutsu specifically, its own read-only host-root mount shadows
  `/dev/null` for reads (matching real host behavior) but rejects writes as a
  read-only-filesystem error, so `cmd > /dev/null` was actively broken, not
  merely missing.
- Just mounting `DevFs` at `/dev` in `with_backend` would have been a no-op:
  `VirtualOverlayBackend::is_virtual_path` hardcoded routing to `/v`/`/v/*`
  only, so any other internal-VFS mount (including a fresh `/dev`) fell
  straight through to the embedder's backend. Added `VfsRouter::has_mount` and
  made `is_virtual_path` mount-aware (keeping the `/v` namespace reservation
  as an explicit fast path, since it's reserved even where nothing is
  mounted), then mounted `/dev` alongside `/v/jobs`/`/v/blobs` in
  `with_backend`.
- Added a regression test (`with_backend_dev_tests.rs`) that builds a
  `with_backend` kernel over a read-only inner backend (kaijutsu's exact
  shape) and verifies `/dev/null` writes discard, reads are empty, and
  `/dev/zero` reads work. Verified it actually catches the bug by reverting
  the two source-side changes and confirming all three cases fail first.

kaijutsu itself is pinned to `kaish-kernel = "0.10"` from crates.io and won't
pick this up until that dependency is bumped after a release — tracked
separately, not part of this PR.

## Test plan

- [x] `cargo test --all`
- [x] `cargo clippy --all --all-targets` — zero warnings
- [x] `cargo check -p kaish-kernel --no-default-features`
- [x] Confirmed new tests fail against the pre-fix source, pass after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -176,6 +176,15 @@ breaking entries are marked **BREAKING**.
   demotion of `\|` to plain `|`).
 
 ### Fixed
+- **`Kernel::with_backend` now mounts `/dev` (DevFs) unconditionally** —
+  custom-backend embedders (e.g. kaijutsu) previously had no `/dev/null`,
+  `/dev/zero`, `/dev/random`, or `/dev/urandom` at all, and `VirtualOverlayBackend`
+  hardcoded routing to the `/v/*` namespace only, so even an embedder that
+  mounted its own `/dev` would have the mount silently ignored. The overlay now
+  also routes any path covered by a real VFS mount (not just `/v/*`), and
+  `with_backend` mounts `/dev` by default — a read-only embedder backend (like
+  kaijutsu's read-only host root) no longer turns `cmd > /dev/null` into a
+  filesystem error.
 - **`${#path}` on an undefined subscripted root is now loud** — `${#nope[items]}`
   silently returned `0` (the bash-parity `${#unset}` → 0 rule leaked into
   subscripted paths), so a typo'd name in a length-guarded loop spun zero times
--- a/crates/kaish-kernel/src/backend/overlay.rs
+++ b/crates/kaish-kernel/src/backend/overlay.rs
@@ -15,8 +15,12 @@
 //!
 //! # Path Routing
 //!
-//! - `/v/*` → Internal VFS (JobFs, MemoryFs for blobs, etc.)
-//! - Everything else → Custom backend
+//! - `/v/*` → always routed to the internal VFS, whether or not anything is
+//!   mounted there (the whole `/v` namespace is reserved, so a miss returns
+//!   `NotFound` from the VFS rather than leaking through to the embedder).
+//! - Any other path actually covered by a mount registered on the internal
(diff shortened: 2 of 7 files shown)


tobert:
## Review (kaibo, cast `deepseek`)

Reviewed the mount-aware `is_virtual_path` logic, the `has_mount`/`find_mount` interaction, backward compatibility for existing `with_backend` embedders, and test coverage.

**Correctness:** No bugs found.
- `VfsRouter::has_mount` (router.rs) is a thin, correct wrapper on the existing `find_mount`, which already prefix-guards with a `/` separator (`/dev` matches `/dev/null` but not `/devnull`).
- The `/v`/`/v/*` fast path in `is_virtual_path` stays a plain string-prefix check (not separator-guarded), but that's safe — the kernel only ever mounts real paths under `/v/*`, never a bare `/v` sibling like `/va`.

**"Does a `/` mount ever make `has_mount` return `true` for everything?"** No — `Kernel::with_backend` only ever mounts `/v/jobs`, `/v/blobs`, and now `/dev`; never `/`. Flagged as a theoretical footgun only if an embedder's own `configure_vfs` closure mounted `/` directly, which would violate the documented contract (mounts belong under `/v`) — a self-inflicted misuse, not a bug in this change. No guard added, matching this project's convention against validating scenarios that can't happen from correct usage.

**Backward compatibility:** Confirmed no behavior change for embedders following the documented `/v/*` convention — those paths are still caught by the string fast path before `has_mount` is ever consulted. For any embedder that previously mounted a non-`/v` path in `configure_vfs`, that mount was previously dead code (silently unreachable); it now works, which is the point of the fix, not a regression.

**Test coverage:** Flagged four gaps, all added in a follow-up commit:
- `ls_dev_lists_devices_under_with_backend` — directory listing through the overlay.
- `non_virtual_write_still_delegates_to_read_only_backend` — proves mount-aware routing didn't widen what counts as internal-VFS-owned; a real non-`/dev` path under the read-only backend still fails.
- `configure_vfs_can_override_dev_mount` — confirms an embedder can still shadow the built-in `/dev` mount, same as it already could for `/v/jobs`/`/v/blobs`.

All 6 tests in `with_backend_dev_tests.rs` pass with the fix, and I verified they're genuine regression tests by reverting the two source files and confirming the write/read/listing cases fail with the exact kaijutsu symptom (read-only-filesystem error on `/dev/null` write) before the fix.

`cargo test --all`, `cargo clippy --all --all-targets` (zero warnings), and `cargo check -p kaish-kernel --no-default-features` are all clean.

PR #82, opened 2026-07-03
fix(kernel): render .data-only $(cmd) as compact JSON in string interpolation

tobert:
## Summary

- Fixes the P3 SILENT-DATA-LOSS item from `docs/issues.md`: quoted command
  substitution (`"...$(cmd)..."`) read a command's result through
  `try_text_out()` alone (`kernel.rs`, `StringPart::CommandSubst`). A
  tool/builtin that sets only structured `.data` with an empty `.out`
  interpolated to `""` — the collection silently evaporated instead of
  rendering or erroring.
- **Decision (Amy):** when `.out` is empty but `.data` is present, render it
  exactly the way a bare `"$x"` collection-valued variable renders — compact
  JSON for lists/records, plain form for scalars — by **reusing
  `value_to_string`**, the same helper `StringPart::Var` already calls. So
  `"$(cmd)"` and `x=$(cmd); "$x"` now display identically; no duplicated
  rendering logic. Non-empty `.out` still wins unchanged, and the
  trailing-newline trim only applies to the text path, not the
  freshly-rendered JSON/scalar string. The existing binary-data loud-error
  branch (`Err(e) => bail!(...)`) is untouched.

## Discovery along the way

Kaish's own `success_data()`-based builtins (`fromjson`, `keys`, `values`)
already populate `.out` with compact JSON, so they never actually hit this
hole in practice today (verified via REPL repro before writing any test).
The genuinely reachable case is a `KernelBackend`-registered embedder tool
returning `ToolResult::with_data("", data)` (or any future core builtin using
`ExecResult::from_parts` directly) — same empty-`.out`/`Some(.data)` shape,
same silent loss. Tests pin that shape directly via fixture `Tool`s registered
through `Kernel::with_backend`, rather than relying on a shipped builtin that
doesn't reproduce it.

**Sync twin checked, not touched:** `scheduler/pipeline.rs::eval_string_parts_sync`
does not have the same hole — its `StringPart::CommandSubst` arm skips
entirely ("requires async"), a separate pre-existing limitation. Left alone
per the note that another agent is working on that function's error
channels.

**Deferred, not fixed here (found but out of scope):** in
`kernel.rs::execute_command_depth`'s backend-tool fallback (~line 2976-2984,
"Try backend-registered tools"), `tool_result.data` is never copied onto the
resulting `ExecResult` at all — an embedder tool's `.data` is dropped outright
in that specific path, distinct from and more severe than the bug this PR
fixes (there `.data` reaches `ExecResult` but renders as `""`; there it never
reaches `ExecResult` in the first place). Worth its own issue/PR.

## Test plan

New file `crates/kaish-kernel/tests/cmdsubst_data_interpolation_tests.rs`,
kernel-routed via `kernel.execute(...)`:
- `record_data_only_result_renders_as_compact_json_in_string` — fixture tool,
  record `.data`, empty `.out` → renders as `{"a":1,"b":2}` (fails before fix).
- `list_data_only_result_renders_as_compact_json_in_string` — same, list shape
  (fails before fix).
- `scalar_data_only_result_renders_unquoted_in_string` — scalar `String`
  `.data` renders unquoted (`hello`, not `"hello"`), proving reuse of
  `value_to_string` rather than a naive JSON stringify (fails before fix).
- `cmdsubst_and_bare_var_render_a_collection_identically` — `"$(cmd)"` vs
  `x=$(cmd); "$x"` produce the same string (fails before fix).
- `nonempty_out_wins_over_data_in_string` — non-empty `.out` always wins over
  `.data` (passes before and after; pins unchanged behavior).
- `fromjson_record_in_string_interpolation` / `keys_list_in_string_interpolation`
  — real shipped builtins through the same path (passes before and after).

Confirmed manually: reverted the `kernel.rs` hunk, the four "fails before fix"
tests failed with the exact `""`-drop symptom (`"parsed:  done"`,
`"items:  done"`, `"say:  done"`, `"got:"`), then restored the fix and all 7
passed.

Gates run clean:
- [x] `cargo test --all`
- [x] `cargo clippy --all --all-targets` (zero warnings)

- [x] `CHANGELOG.md` — one bullet under `## [Unreleased]` → `### Fixed`

Not merging — leaving for review per repo convention (kaibo/code-review before merge).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -193,6 +193,13 @@ breaking entries are marked **BREAKING**.
   demotion of `\|` to plain `|`).
 
 ### Fixed
+- **`"$(cmd)"` no longer drops a `.data`-only result to `""`** — quoted command
+  substitution used to read only a command's `.out` text; a tool that set
+  structured `.data` (a list/record) but left `.out` empty interpolated to
+  nothing, silently losing the collection. It now falls back to rendering
+  `.data` the same way a bare `"$x"` renders a collection-valued variable
+  (compact JSON for lists/records, plain form for scalars); non-empty `.out`
+  still wins unchanged.
 - **`Kernel::with_backend` now mounts `/dev` (DevFs) unconditionally** —
   custom-backend embedders (e.g. kaijutsu) previously had no `/dev/null`,
   `/dev/zero`, `/dev/random`, or `/dev/urandom` at all, and `VirtualOverlayBackend`
--- a/crates/kaish-kernel/src/kernel.rs
+++ b/crates/kaish-kernel/src/kernel.rs
@@ -4140,7 +4140,22 @@ impl Kernel {
                 // Embedding binary into a string is a text context: fail loud
                 // rather than splice in U+FFFD garbage.
                 match result.try_text_out() {
-                    Ok(s) => Ok(s.trim_end_matches('\n').to_string()),
+                    // Text wins when present — unchanged behavior.
+                    Ok(s) if !s.is_empty() => Ok(s.trim_end_matches('\n').to_string()),
+                    // `.out` is empty: a builtin/tool that set only structured
+                    // `.data` must not silently evaporate to "" (SILENT DATA
+                    // LOSS). Render it the same way a bare `"$x"`
+                    // collection-valued variable renders — compact JSON for
+                    // lists/records, plain form for scalars — by reusing
+                    // `value_to_string` (the exact `StringPart::Var` helper
+                    // above) so `"$(cmd)"` and `x=$(cmd); "$x"` display
+                    // identically. No trailing-newline trim here: that's a
+                    // text-path artifact, not applicable to a freshly
+                    // rendered JSON/scalar string.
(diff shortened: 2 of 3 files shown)


tobert:
kaibo review (deepseek, grounded against the branch worktree): **clean**.

- All three arms of the `CommandSubst` match verified correct: non-empty `.out` wins with the POSIX trailing-newline trim (text path only); empty-`.out`+`.data` renders via `value_to_string` — the identical helper the `StringPart::Var` arm uses, pinned by the scalar test proving unquoted rendering; binary loud-error unchanged.
- No codebase pattern exists where empty `.out` + incidental `.data` should NOT render — `success_data()` always materializes `.out`, so the only producer of the bug shape is an explicit structured-data-only result.
- All 7 tests route through `kernel.execute` per the contributor convention.
- The deferred discovery in the PR body is confirmed real and now filed as #84 (backend-tool dispatch drops `ToolResult.data`/`content_type`/`baggage`; verified by direct read at kernel.rs:2976-2984).

PR #83, opened 2026-07-03
docs(help): document collections known limitations for 0.11.0

tobert:
## Summary

`docs/issues.md`'s "Before 0.11.0" section flagged three loud gaps on the
collections surface to either fix or ship as documented known limitations
before release. Per the task, I verified each empirically against `main`
(REPL runs + existing test coverage) before writing anything down — and only
one turned out to still be a live gap:

- **Bracket-path `push` target** (`push services[web][tags] item`) —
  **confirmed, still broken.** It fuses into a glob word and fails loudly
  ("no matches") before `push` runs. Documented in `help collections`
  (`crates/kaish-help/src/fragments.rs`, regenerated `syntax.md`) and in a new
  `arrays-and-hashes.md` "Known limitations (0.11.0)" section, with the
  verified workaround: read the nested list into a temp var, `push`, assign
  it back. The actual fix stays open under `docs/issues.md` P3.
- **Deeply-nested glued list literals** (`x=[[a] [b]]`) — **already fixed.**
  Traced to commit `09c1a89` (the `[[ ]]` test-depth rewrite of
  `lexer::compute_value_context`), which fixed this as an unintentional side
  effect the same day the original deferral was written — nobody credited it.
  Added a regression test (`deeply_nested_glued_list_literal_matches_spaced_nesting`)
  pinning glued vs. spaced nesting at depth 2 and 3, corrected the stale
  comment above `struct ValueContext` in `lexer.rs`, and corrected the stale
  claims in `arrays-and-hashes.md` / `docs/issues.md`.
- **Nested `${#path}` / `${path:-default}`** — **already fixed.** Shipped in
  `28ec480` before the "Before 0.11.0" list was even written, and already
  fully covered by `collection_access_tests.rs`. While verifying this I found
  one adjacent, previously-undocumented gap: `${#var}` length syntax (nested
  or not) isn't recognized at all inside `$(( ))` arithmetic — recorded as a
  new P4 bullet in `docs/issues.md` rather than silently dropped.

Corrected the `docs/issues.md` entries that described the two fixed items as
still-open (the P2 "STRUCTURAL #6 resolver" entry's "folds in" clause and
"riders" list, and the P3 "Collection literals" entry), so the backlog now
reflects only real open work: the one documented `push` limitation, the
reduced-sync silent-coalesce items, and the `"$(cmd)"` data-loss item.

## Test plan

- [x] `cargo test --all` — clean (104 test binaries/doctests, all `ok`,
      including the `kaish-help` drift test `syntax_md_matches_fragments` and
      the new `collection_literals_tests::deeply_nested_glued_list_literal_matches_spaced_nesting`)
- [x] `cargo clippy --all --all-targets` — zero warnings, zero errors
- [x] `cargo run -p kaish-help --example regen_syntax` — regenerated
      `content/en/syntax.md`, no drift
- [x] Manually verified `help collections` renders the new limitation note
      end-to-end via `kaish-repl -c 'help collections'`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,12 @@ breaking entries are marked **BREAKING**.
   stale since list/record literals landed. Membership (`in`/`not in`) is now
   always shown inside a full `if [[ … ]]; then … fi`, never a bare standalone
   test line (a prior regression against the shell's own teaching-note rule).
+- **`help collections` documents the one remaining 0.11.0 known limitation** —
+  `push` only accepts a top-level bareword target; a bracket-path target
+  (`push services[web][tags] item`) fails loudly instead of appending, with the
+  read/push/assign-back workaround shown inline. Two other suspected gaps
+  (deeply-nested glued list literals, nested `${#path}`/`${path:-default}`)
+  turned out already fixed on re-verification and needed no documentation.
 
 ### Added
 - **`help regex`** — a compact syntax section (also in `help syntax`/`syntax.md`)
--- a/crates/kaish-help/content/en/syntax.md
+++ b/crates/kaish-help/content/en/syntax.md
@@ -106,6 +106,14 @@ user.email=x              # error — brackets only, use `user[email]=x`
 # push appends to a top-level LIST in place — takes the NAME (like read/unset)
 push xs date               # xs is now [...,"date"]
 push xs $rec               # values push as typed Values, not stringified text
+
+# KNOWN LIMITATION — push only takes a top-level bareword target; a bracket
+# path target fuses into a glob and fails loudly ("no matches") instead of
+# pushing. Read the nested list out, push, assign it back:
+push services[web][tags] item   # ERROR — no matches: services[web][tags]
+tmp=${services[web][tags]}      # workaround
+push tmp item
+services[web][tags]=$tmp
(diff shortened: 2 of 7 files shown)


tobert:
kaibo review (deepseek, grounded against the branch worktree): **clean on all four verification points**.

- The documented `push` workaround (`tmp=${services[web][tags]}; push tmp item; services[web][tags]=$tmp`) is valid kaish on this branch — verified against `resolve_step`, `walk_append` (push.rs:110 → scope.rs:843), and `walk_write` subscripted assignment.
- The corrected `ValueContext` comment accurately describes why glued nested literals parse: the `!currently_value` guard on `[[` detection (lexer.rs:100-105) routes a nested `[` to `bracket_depth`, never `test_depth`.
- Regression-test expectations (`[["a"],["b"]]`) match the established echo rendering convention in the same test file.
- No factual errors in the arrays-and-hashes.md "Known limitations (0.11.0)" section or the issues.md corrections; the `09c1a89`/`28ec480` fix attributions check out.

Note: heads-up for merge ordering — this PR and #85 both edit `docs/issues.md` and `CHANGELOG.md` (Unreleased); expect a trivial conflict in whichever lands second.

PR #85, opened 2026-07-03
fix(scheduler): loud errors for collection path failures in the sync eval path

tobert:
## Summary

0.11.0 pre-release punch-list item (docs/issues.md, now closed): `scatter`/`gather`
bind their own flag values (`--as`, `--limit`, `--timeout`) through a *sync* twin
of the real async argument binder — `build_tool_args`/`eval_simple_expr` in
`scheduler/pipeline.rs` — because `run_scatter_gather` parses those flags once,
before any worker forks, and can't recurse back through the async
`PipelineRunner::run` → `Kernel::dispatch` chain to get there.

Every real command's arguments bind through the async `build_args_async`
(`kernel.rs`), which already surfaces a `PathError` (missing key, shape
mismatch, undefined subscripted root) loudly — matching the other three
primary eval sites (assignment, `$(( ))`, `"${…}"`). The sync twin never got
that treatment: every arm discarded the error via `.ok()`/`if let Ok(..)`, so
`scatter --as ${u[nope]}` silently fell back to treating `--as` as a bare
boolean flag (dropping the intended value and stranding the bad expression as
an unused positional) instead of failing, and `${#u[tags]}` on a bad subscript
inside an interpolated flag value silently omitted the length instead of
erroring.

## What changed

- `build_tool_args` is now `Result<ToolArgs, String>`; `eval_simple_expr` is
  now `Result<Option<Value>, String>`; `eval_string_parts_sync` is now
  `Result<String, String>`. `Ok(None)` still means "not representable in this
  reduced sync context" (binary ops, command substitution — the documented,
  deferred "eliminate the sync twin" gap); `Err(msg)` is a genuine `PathError`
  (absence/shape), propagated with the same message text the async path uses.
  An unset *bare* root still coalesces (bash-compatible), unchanged.
- `run_scatter_gather`'s two `build_tool_args` calls now convert an `Err` into
  `ExecResult::failure(1, …)` — the same architectural boundary `run_single`
  already uses for a dispatch error.
- Added two previously-missing `eval_simple_expr` match arms for *bare*
  (unquoted whole-token) `Expr::VarLength`/`Expr::VarWithDefault` —
  `scatter --limit ${#tags}` used to fall through to the catch-all `_ => None`
  (silently dropping the flag entirely). They now route through the shared
  `resolve_length`/`resolve_default` the async path already calls.
- Deduped `eval_simple_expr`'s `Expr::Interpolated` arm, which had its own
  ~60-line copy of `eval_string_parts_sync`'s loop, down to a single call.
- Updated the test-only `BackendDispatcher` (`dispatch.rs`) and its ~19
  in-file unit tests in `pipeline.rs` to the new fallible signature.

Confirmed `build_tool_args` has exactly one production call site — the two
calls in `run_scatter_gather` — everywhere else is `#[cfg(test)]` or
`pipeline.rs`'s own unit tests, so this closes the silent-corruption surface
without touching the deferred larger "eliminate the sync twin" refactor.

## Decisions

- Kept `PathError::UndefinedRoot` on a *bare* variable coalescing to
  empty/skip (matches bash and the async path's string-interpolation
  convention) — only `Absence`/`Shape` on a *subscripted* path are loud, same
  split the async path and `interpreter/eval.rs` already use.
- Left the parallel `StringPart::Arithmetic` swallow in the same two
  functions alone — same symptom class (a bad `scatter --limit $((1/0))`
  would silently omit the digits), but a distinct `anyhow` error type, not
  `PathError` — out of this fix's scope. Recorded in `docs/issues.md` for the
  next pass.

## Test plan

- [x] TDD: wrote the new kernel-routed tests in
      `crates/kaish-kernel/tests/scatter_gather_jsonl_tests.rs` against the
      unmodified source first; confirmed 2 failed as expected (a bad
      subscript and a shape error in `--as` both silently exited 0). A third
      test needed a second pass — its worker read `$N` while the buggy
      fallback actually bound `$n`, so it was accidentally passing "for free"
      for an unrelated reason; fixed to read `$n` and reconfirmed red on old
      code, green after the fix.
- [x] `cargo test --all` — clean
- [x] `cargo clippy --all --all-targets` — zero warnings
- [x] `cargo check -p kaish-kernel --no-default-features` — clean
- [x] kaibo review (deepseek cast) — couldn't check out the branch (known
      sandbox limitation), so it did a predictive semantic-trace review
      against `main` + this PR's description instead of a live diff; every
      item it flagged as "must verify" (UndefinedRoot/Absence/Shape split,
      `.map_err`/anyhow wrapping in `dispatch.rs`, `?` placement, `Ok(...)`
      wrapping, the `PathError` import, test `.expect()` calls, error-string
      preservation in `run_scatter_gather`) checks out against the actual diff.
- [x] `CHANGELOG.md` bullet under `Fixed`; `docs/issues.md` entries closed;
      `docs/devlog.md` entry added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -237,6 +237,13 @@ breaking entries are marked **BREAKING**.
     default on a *subscripted* path — `${#u[tags]}`, `${cfg[port]:-8080}` — were
     briefly a loud "bind it first" error; they are now fully path-aware, see
     Added.)
+- **`scatter`/`gather`'s own flag values (`--as`, `--limit`, …) now fail loud on
+  a bad/subscripted collection access** — the reduced sync arg binder they use
+  (it runs once, before workers fork, so it can't recurse through the async
+  pipeline) used to discard a missing-key/shape `PathError` as a silently
+  dropped flag or an omitted `${#…}` length instead of erroring, unlike `echo`,
+  assignment, `$(( ))`, and `"${…}"`. A bad `scatter --as ${u[nope]}` now fails
+  the whole pipeline instead of quietly falling back to a bare boolean flag.
 - **A bare collection reaching an external command's argv or a redirect
   target silently JSON-serialized** (e.g. `curl -d $cfg` or `echo x > $cfg`);
   every process-boundary sink now refuses it with a `$(tojson $x)` hint. This
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -243,7 +243,9 @@ breaking entries are marked **BREAKING**.
   pipeline) used to discard a missing-key/shape `PathError` as a silently
   dropped flag or an omitted `${#…}` length instead of erroring, unlike `echo`,
   assignment, `$(( ))`, and `"${…}"`. A bad `scatter --as ${u[nope]}` now fails
-  the whole pipeline instead of quietly falling back to a bare boolean flag.
+  the whole pipeline instead of quietly falling back to a bare boolean flag; a
+  subscripted path on an entirely undefined root (`${nope[key]}`) is loud too,
+  while a bare undefined `$VAR` keeps the bash-compatible coalesce.
 - **A bare collection reaching an external command's argv or a redirect
   target silently JSON-serialized** (e.g. `curl -d $cfg` or `echo x > $cfg`);
   every process-boundary sink now refuses it with a `$(tojson $x)` hint. This
--- a/crates/kaish-kernel/src/dispatch.rs
+++ b/crates/kaish-kernel/src/dispatch.rs
@@ -427,9 +427,13 @@ impl CommandDispatcher for BackendDispatcher {
             _ => {}
         }
 
-        // Build tool args with schema-aware parsing (sync — no command substitution)
+        // Build tool args with schema-aware parsing (sync — no command substitution).
+        // A bad/subscripted collection access is a genuine PathError here too —
+        // propagate it via `?` rather than swallowing, same as the production
+        // Kernel::dispatch_command's `execute_command(..).await?`.
         let schema = self.tools.get(&cmd.name).map(|t| t.schema());
-        let tool_args = build_tool_args(&cmd.args, ctx, schema.as_ref());
+        let tool_args = build_tool_args(&cmd.args, ctx, schema.as_ref())
+            .map_err(|e| anyhow::anyhow!(e))?;
 
         // Honor --json before the tool runs so a parse failure inside the
         // builtin doesn't drop the format on the floor. See kernel.rs for the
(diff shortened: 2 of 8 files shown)


tobert:
**Review delta (kaibo, deepseek cast — approved with one finding + one test-gap check):**

**Finding applied** (`a5cdb76`): `eval_simple_expr`'s `Expr::VarRef` arm coalesced `PathError::UndefinedRoot` to `Ok(None)` *unconditionally*, so `scatter --as ${x[key]}` with a typo'd (entirely undefined) root still silently dropped the flag — an undefined root is `UndefinedRoot`, not the `Absence` the first cut made loud. The coalesce is now restricted to bare paths (`path.segments.len() <= 1`), the same split `resolve_length` draws; a subscripted path on an undefined root errors with the same `"${x[key]}: undefined variable"` shape (`format_path` exported `pub(crate)` for it). New tests: `undefined_root_subscript_in_scatter_flag_value_is_loud` (red-checked against the previous commit — it silently exited 0 before the fix) and `bare_undefined_var_in_scatter_flag_still_coalesces` (the bash-compatible bare case, unchanged).

**Deliberate non-change:** did NOT apply the bare-only restriction to `eval_string_parts_sync`'s `StringPart::Var` arm, though it has the same pattern. Verified empirically that both primary string-interpolation sites (async `eval_string_part_async`, sync `eval_interpolated`) expand an undefined root to empty *even when subscripted* — `echo "a${nope[k]}b"` → `ab`, exit 0, bash-compatible — so restricting only the sync twin would have made it stricter than the four primary sites, diverging the other way. The shipped contract is a whole-token vs. string-context split (`echo ${nope[k]}` loud, `"${nope[k]}"` empty); the sync path now matches it on both sides, with a comment on the arm citing the evidence.

**Test gap closed:** the bare (unquoted whole-token) `Expr::VarLength`/`Expr::VarWithDefault` arms are genuinely reachable — the parser emits both variants in expression position (`parser.rs:2307` / `parser.rs:48`) — and are now pinned by three observable tests (`bare_length_bad_subscript_in_scatter_flag_is_loud` as the reachability proof: a regression to the old `_ => None` catch-all makes that pipeline silently succeed; `bare_length_in_scatter_flag_value_succeeds`; `bare_default_in_scatter_flag_value_binds_the_var_name`, which fails loud at 123 if the arm regresses because the dropped `--as` leaves the ITEM binding in place).

Gates re-run clean: `cargo test --all` (104 test binaries, 0 failures) and `cargo clippy --all --all-targets` (0 warnings).

tobert:
Superseded by #87 — identical content rebased onto main (post #82/#83); the force-push guard makes a fresh branch the convention. Review record stays here.

PR #86, opened 2026-07-03
feat(#80): fromjsonl/tojsonl JSONL doors; jq gets real -s and hint errors

### LINKED ISSUE #80: jq: document-stream semantics for multi-document (JSONL) text input
**Fix before 0.11.0** — softens the `.[]` watch item from the scatter/gather redesign (#73 / PR #79).

> **Revision note:** this issue originally proposed jq auto-slurping multi-document text. Design review (2026-07-03) worked back from scatter/gather needs and landed on **explicit named doors instead of implicit format handling**: a `fromjsonl`/`tojsonl` builtin pair (matching `fromjson`/`tojson`), with jq staying strictly one-document and getting louder, hint-carrying errors. Original text in the edit history.

## Why (worked back from scatter/gather)

1. **External JSONL → typed scatter is impossible today.** On the #79 branch, scatter's text ingress deliberately binds each line as a *string* (`ScatterItem::from_text_line`, no JSON-sniffing). Typed items only come from `.data`, and both text→typed bridges (`jq`, `fromjson`) hard-reject multi-document text. There is no path from `curl` of an NDJSON API or a log file to `${ITEM[id]}` workers.
2. **kaish can't re-read its own gather output.** #79 makes JSONL the default gather egress — the lingua franca — but `… | gather > results.jsonl` produces a file kaish's own jq cannot ingest. Egress door, no ingress door.
3. **Per-document jq streaming considered and rejected.** Real-jq per-doc semantics would make framing depend on transport (the same rows need `.[]` in-pipe via `.data` but not via text — two idioms selected by an invisible property of the upstream stage) and break aggregation uniformity (`jq add`, `map | length` over a stream). The always-slurped one-document model stays: *jq gets exactly one document; plural is an array.*

## `fromjsonl` — JSONL text → typed list

Ingress half of the stream boundary. Mirrors `fromjson` conventions (string positional or piped stdin text; envelope-free conversion; `.data` out).

- **Strictly line-oriented** (the name is the contract): each non-blank line parses as exactly one complete JSON document → element of the output list. Multi-line pretty-printed documents are a loud error (that's `fromjson` / `jq -s` territory).
- Blank lines skipped (same panel-ratified rule as scatter ingress).
- Any unparseable line — including a trailing partial line — is a **loud error with the line number**, never a silent skip.
- `null` elements are allowed (it's data; scatter guards nulls at its own boundary with its existing loud error).
- Empty input → empty list, exit 0 (a stream of zero documents is a legitimate stream — deliberate asymmetry with `fromjson`, where empty input stays loud).

Canonical flow this unlocks: `curl …ndjson | fromjsonl | scatter -j8 -- restart ${ITEM[host]}` and `cat results.jsonl | fromjsonl | jq '.[] | select(.ok | not)'`.

## `tojsonl` — list → JSONL text

Egress half. Mirrors `tojson` conventions (one Value positional or piped `.data`; **text out, no `.data`** — the point is the text form; `Value::Bytes` loud).

- Input must be a **list**: each element renders as one compact JSON document per line, newline-terminated.
- Non-list input is a loud error with the fix in the message (“a record/scalar is one document — use `tojson`, or wrap it in a list”).
- Always compact, no `--pretty` (pretty-printed JSONL isn't a thing).
- Empty list → empty output, exit 0 (round-trips with `fromjsonl`).

## jq — stays one-document, gets louder and real-jq `-s`

- **Default text path unchanged** (one document), but the multi-doc parse error gets smart: on “trailing characters”, attempt a stream parse; if the remainder is more JSON documents, say so — `jq: input looks like JSONL (N documents) — kaish jq takes one document; use fromjsonl upstream, or jq -s`. Garbage stays the plain loud error.
- **Accept `-s`/`--slurp` with real jq semantics on the text path** — parse the document stream (whitespace-separated, real-jq framing, so pretty multi-doc works) and *always* wrap in an array, single doc included (`[doc]`). NOT a no-op: no-op `-s` silently diverges on single-doc input (`printf '{"a":1}' | jq -s length` must be `1`, not the key count). On the `.data` path `-s` IS a no-op — the stream is already slurped, so `gather | jq -s '.[] | …'` matches real-jq muscle memory exactly.
- **`.[]` hint rider** (the actual #79 watch item): the runtime error `Cannot index array with "…"` is the exact signature of writing the real-jq streaming form against a slurped array — append `kaish jq is always-slurped — try '.[] | …'`. If field evidence shows agents thrashing past the hint, the revisit is an opt-in flag, never a default change.

## Tests (kernel-routed)

- `fromjsonl`: happy path typed list, blank-line skip, partial trailing line loud (with line number), garbage line loud, empty input → `[]`, null element passes, multi-line pretty doc loud.
- `tojsonl`: list → lines, non-list loud, empty list → empty/0, `tojsonl` ∘ `fromjsonl` round-trip.
- End-to-end: `gather > file` then `cat file | fromjsonl | scatter` re-fans out **typed** (`${ITEM[…]}` subscripts, `typeof` guards).
- jq: single-doc text unchanged; JSONL-shaped input → hint error; `-s` single-doc wrap + multi-doc slurp + `.data` no-op; `Cannot index array` hint.

## Docs

Help fragments for both builtins; `help scatter` / `help collections` cross-reference the JSONL doors; `docs/LANGUAGE.md` text-boundary section; CHANGELOG `Added` (fromjsonl, tojsonl) + `Changed` (jq `-s`, hint errors).

tobert:
## Summary

Closes #80. Worked back from scatter/gather (#79)'s own watch item: `gather`
made JSONL the default egress, but kaish had no path back in — `fromjson`/`jq`
both hard-reject multi-document text, so an NDJSON API stream, a log file, or
even kaish's own `gather > file` output had no way to re-fan-out typed.

**Decisions** (from the issue's design-review revision note, 2026-07-03):

- **Explicit named doors over implicit slurp.** Per-document jq streaming was
  considered and rejected — it would make framing depend on an invisible
  property of the upstream stage (`.[]` in-pipe via `.data` but not via text)
  and break aggregation uniformity (`jq add`, `map | length`). Landed on
  `fromjsonl` (ingress: JSONL text → typed list) / `tojsonl` (egress: list →
  JSONL text) as the explicit pair, mirroring `fromjson`/`tojson`.
- **jq stays strictly one-document.** `jq` unchanged on the default text path;
  `-s`/`--slurp` gets *real* jq semantics (parses the whole document stream,
  always wraps in an array — even a single document, so it doesn't silently
  diverge on `printf '{"a":1}' | jq -s length` → must be `1`, not the key
  count). No-op on the `.data` path (already one structured value there).
- **Louder, not smarter.** A "trailing characters" parse error now checks
  whether the remainder is actually more JSON documents and, if so, names the
  count and points at `fromjsonl`/`jq -s`; genuine garbage keeps the plain
  error. The runtime `cannot index [...] with ...` error — the actual #79
  watch item (real-jq's per-row form against kaish's always-slurped array) —
  gains a `.[] | …` hint, per the issue's specified wording.
- **`fromjsonl` is strictly line-oriented** (the name is the contract): blank
  lines skipped, any bad line including a truncated trailing one is loud with
  the line number, empty input → `[]` (a legitimate zero-document stream —
  falls out naturally from the blank-line rule, no special case needed). A
  pretty-printed multi-line document is diagnosed specifically and points at
  `fromjson`/`jq -s` instead of a bare per-line error.
- **`tojsonl` also accepts piped `.data`** (unlike `tojson`, which is
  positional-only) so it composes as a real pipeline stage
  (`curl … | fromjson | tojsonl`), not just a `$()`-capture step.

**Bug found and fixed along the way:** building the kernel-routed
`gather > file; cat file | fromjsonl | scatter` round trip surfaced that the
scatter/gather special-cased pipeline path never applied `gather`'s own
trailing redirect when gather was the pipeline's last command
(`ScatterGatherRunner` never saw `gather_cmd.redirects`, only the parsed
`--json`/`--lines` flags) — the JSONL rows landed back in the result instead
of the file. Fixed in `scheduler/pipeline.rs`'s `run_scatter_gather`.

**Known, deferred edge case:** a redirect on `gather` combined with
`post_gather` pipeline stages (e.g. `… | gather > file | jq …`, redirect in
the middle of a pipe) still silently drops the redirect — pre-existing
behavior, unchanged by this PR, not part of #80's scope. Worth a follow-up if
it ever bites someone in practice.

## Test plan

- [x] `cargo test --all` — full suite clean, including 21 new kernel-routed
      cases in `jsonl_doors_tests.rs` (fromjsonl happy/blank-skip/truncated/
      garbage/empty/null/pretty-multiline; tojsonl list/non-list/empty/
      round-trip; jq single-doc/JSONL-hint/-s-wrap/-s-slurp/-s-data-noop/
      index-hint; the `gather > file | fromjsonl | scatter` round trip) plus
      2 new `--json` sweep cases.
- [x] `cargo clippy --all --all-targets` — zero warnings.
- [x] `cargo check -p kaish-kernel --no-default-features` — pure-data pair
      compiles and the new tests run without `localfs`.
- [x] `cargo insta test --check` — no pending snapshots; `syntax.md`
      regenerated after the collections fragment's one-line cross-reference.
- [x] Independent review via kaibo (deepseek cast) — flagged the
      pipeline.rs fix and jq_native.rs helpers as sound; other findings were
      either false positives (verified `split_inclusive` handles a
      non-newline-terminated tail correctly) or wording dictated verbatim by
      the ratified issue spec.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,28 @@ breaking entries are marked **BREAKING**.
 
 ## [Unreleased]
 
+### Added
+- **`fromjsonl` / `tojsonl`** — the JSONL (NDJSON) doors, GH #80. `fromjsonl`
+  parses strictly line-oriented JSONL text (one JSON document per line, blank
+  lines skipped, any bad line — including a truncated trailing one — a loud
+  error naming the line number; empty input is a legitimate zero-document
+  stream → `[]`) into a typed list in `.data`. `tojsonl` is the egress mirror:
+  a list in, one compact JSON document per line out (a record/scalar is one
+  document — that's `tojson`'s job). Unlocks `curl … | fromjsonl | scatter`
+  and re-ingesting kaish's own `gather` output (`cat results.jsonl | fromjsonl
+  | jq …`).
+
 ### Changed
+- **`jq -s`/`--slurp` is now real jq slurp semantics**, not a no-op (GH #80).
+  On the text path it parses the whole whitespace-separated document stream
+  and always wraps the result in an array — even a single document
+  (`printf '{"a":1}' | jq -s length` → `1`, the array length, not a no-op's
+  accidental key count). On the `.data` pipeline path `-s` stays a no-op (the
+  upstream stage already handed over one structured value). Plain `jq` (no
+  `-s`) on JSONL-shaped text now names the document count and points at
+  `fromjsonl`/`jq -s` instead of a bare "trailing characters" parse error; the
+  runtime `cannot index [...] with ...` error gains a `.[] | …` hint (kaish
+  jq is always-slurped).
 - **BREAKING: scatter/gather is now the typed parallel map** (GH #73,
   panel-validated). `gather` emits one JSONL result record per worker, in item
   order, failures included — `{"i":N,"item":<typed>,"ok":…,"code":…,"out":…,
@@ -193,6 +214,11 @@ breaking entries are marked **BREAKING**.
   demotion of `\|` to plain `|`).
 
 ### Fixed
+- **`gather`'s own trailing redirect (`… | gather > file`) is no longer silently
+  dropped.** Found while wiring up the JSONL doors (GH #80): the scatter/gather
+  special-cased pipeline path never applied `gather`'s redirects when it was
+  the pipeline's last command — the JSONL rows landed back in the result
+  instead of the file.
 - **`Kernel::with_backend` now mounts `/dev` (DevFs) unconditionally** —
   custom-backend embedders (e.g. kaijutsu) previously had no `/dev/null`,
   `/dev/zero`, `/dev/random`, or `/dev/urandom` at all, and `VirtualOverlayBackend`
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ everywhere, a `jq` that always uses the same filter syntax, an `awk` that never 
 |----------|-------|
 | **Text** | awk, base64, cut, diff, grep, head, sed, sort, split, tac, tail, tr, uniq, wc, xxd |
 | **Files** | basename, cat, cd, checksum, cmp, cp, dd, dirname, file, find, glob, ln, ls, mkdir, mktemp, mv, patch, pwd, readlink, realpath, rm, stat, tee, touch, tree, write |
-| **JSON** | fromjson, jq, keys, tojson, typeof, values |
+| **JSON** | fromjson, fromjsonl, jq, keys, tojson, tojsonl, typeof, values |
 | **System** | alias, bg, date, echo, env, exec, export, fg, help, hostname, jobs, kill, printf, ps, push, read, seq, set, sleep, spawn, timeout, tokens, uname, unalias, unset, wait, which |
 | **Parallel** | scatter, gather |
 | **Meta** | assert, false, true |
--- a/crates/kaish-help/content/en/scatter.md
+++ b/crates/kaish-help/content/en/scatter.md
@@ -28,7 +28,9 @@ worker binds the real element — `${ITEM[id]}` subscripts a record, numbers sta
 numbers, `1` ≠ `"1"`. From plain text, one string item per line (blank lines
 skipped). Quote it at command boundaries: `work "$ITEM"` (a record renders as
 compact JSON). A single non-array object, a `null` element, or binary input is
-a loud error.
+a loud error. External JSONL (an API stream, a log file) fans out **typed**
+only through `fromjsonl` first — `curl … | fromjsonl | scatter …` — see
+`help fromjsonl`.
 
 ## Consuming results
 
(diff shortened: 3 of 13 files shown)


tobert:
Addressed the kaibo (deepseek) review findings in 9461b57:

- **Test: `tojsonl` bytes rejection** — `tojsonl_bytes_is_loud_never_an_envelope` pins the loud "binary — use base64" error and asserts no base64 envelope leaks to stdout.
- **Test: embedded-newline safety** — `tojsonl_embedded_newline_stays_one_line` pins the central contract: a string element containing a real newline serializes as exactly one output line with `\n` escaped, and re-ingests through `fromjsonl` unchanged.
- **Test: CRLF line endings** — `fromjsonl_accepts_crlf_line_endings` documents the split-on-`\n`-then-strip-one-`\r` contract.
- **Doc fix** — fromjsonl's already-structured-positional comment no longer claims idempotence; `fromjsonl $v` yields `[$v]` (one-element wrap), and the comment now says so.

`cargo test --all` (105 binaries) and `cargo clippy --all --all-targets` both clean. Holding off on the rebase onto moved main until #85 (which also touches `run_scatter_gather`) lands, per coordination.

tobert:
Superseded by #89 — identical content rebased onto main past #82/#83/#87 (force-push guard convention: fresh rebased branch + superseding PR instead of a history rewrite). The kaibo review and addressed-findings record stays here.

Issue #93, opened 2026-07-04
0.11.1 residual hardening: binary-at-remaining-text-sinks + review tail


tobert:
Residual hardening items deferred from the 0.11.0 pre-release AI-review pass (two rounds of DeepSeek + Gemini Pro review of the scatter/gather + collections + interpreter surface). None are release-blockers for 0.11.0 — the confirmed silent-corruption/panic bugs were fixed on the `fix/0.11.0-prerelease-hardening` branch. These are the lower-severity / broader-scope tails, grouped for 0.11.1.

## 1. Binary (`Value::Bytes`) still coerces silently at the *remaining* text sinks

0.11.0 made `Value::Bytes` go **loud** at the primary text sinks — string interpolation (`"x=$b"`), bare-word external argv (`prog $b`), and `echo` — via the new fallible `value_to_text_sink()`. This is a **live** bug class (not latent): `cat`/`head`/`tail`/`base64 -d`/`xxd -r`/`dd`/`tee` already emit `OutputPayload::Bytes` for non-UTF-8 content. The following sinks still silently coerce a binary value (via the infallible `value_to_string` → `[binary: N bytes]` placeholder, or a lossy stringify) and should adopt the same loud guard:

- **Builtin path-positional coercion** — `mkdir $BIN`, `cp $BIN dst`, `rm $BIN`, etc. read a typed `Value::Bytes` positional and `value_to_string` it into a path.
- **Env-var export** — `kernel.rs` `cmd.env(name, value_to_string(&value))` (external spawn) and `env`/`export` paths.
- **Redirect target** — `pipeline.rs:176` stringifies a `Value` redirect target; collections are already guarded, binary is not.
- **Heredoc body & record-key interpolation** — the sync heredoc-body push and `RecordKey::Interpolated` assembly.
- **Semantic ops** — `==`, `in`, `${#…}`, `case`-glob on a `Value::Bytes` operand (currently stringify via `value_to_string`; arguably should be a loud type error, matching the Decision-E shape guards).

Also covers the two related SUSPECTED review findings: **S3** (`Value::Bytes` placeholder leaking into external argv — mostly closed by the bare-word fix, residual is the typed-positional-to-builtin path) and **S4** (binary stdin into `fromjsonl`/`fromjson` — verify these reject/handle binary input loudly).

The clean approach is to keep pushing `value_to_text_sink()` (or `try_text_out`-style guards) out to these sinks, leaving `value_to_string` only for genuinely-internal/debug uses.

[...]

PR #94, opened 2026-07-04
fix(0.11.0): pre-release hardening — loud errors over silent corruption (2 AI-review rounds)

tobert:
## Why

Before cutting 0.11.0 we hardened the release surface (collections + scatter/gather + the kernel↔backend seam + the interpreter) against silent data corruption, in the release's own spirit: **loud errors, never silent fallbacks; crash beats corrupt.** The work came out of **two rounds of independent, cross-model AI review**:

- **Round 1** — overnight batch reviews (Anthropic "fable" + Gemini Pro) of everything shipped since v0.10.0. Landed as the first commit here.
- **Round 2** — fresh DeepSeek + Gemini Pro reviews of *round 1's own fixes* and the surrounding scatter/gather + interpreter code. The two models were genuinely complementary — each caught real defects the other missed.

Every finding was **verified against the code before trusting it** (a round-1 lesson: one "silent corruption" claim turned out to be the deliberate, tested `kaish-last` feature). Notably, two scary-sounding claims were investigated and found **already-correct**, so left untouched: `gather --json` "bypassing the error envelope" (it deliberately owns its output; per-row `ok`/`code`/`err` *is* the error surface) and `scatter --timeout` "not cancelling a hung worker" (the timer → `worker_token` → `ctx.cancel` → `try_execute_external` chain is intact).

## What's fixed

**scatter / gather (scheduler):**
- Wrong-typed `--limit`/`--as`/`--timeout` values (a string `--limit "5"` from a variable, a negative `--timeout`, a `$(…)` the reduced sync binder can't evaluate) **fail loud** instead of silently falling back to the default. Numeric-string `--limit "5"` coerces.
- Removed the dead `--progress` flag (parsed into an unread field, not even in the clap schema).
- Binary worker output no longer corrupts `gather`: rows and `--lines` used to lossily UTF-8-decode (`U+FFFD`) a worker whose stdout was binary — now a row is `ok:false` with a clear error and `--lines` hard-errors (exit 123). **Live bug**, not latent — `cat`/`head`/`tail`/`base64 -d`/`xxd -r`/`dd`/`tee` already emit binary output.
- Workers now derive their context from `child_for_pipeline()` instead of a from-scratch `ExecContext`, so they keep the parent's **script watchdog** (a from-scratch `None` was clobbering the fork's watchdog → `ctx.patient` holds suspended a missing clock → false-positive request-timeout kills) and honour the **output-limit memory cap** (added a per-worker spill boundary; N parallel workers could each hold unbounded output).

**interpreter — binary at text sinks:** the live async `$()` capture preserves binary as `Value::Bytes`, so `b=$(cat blob)` is genuinely binary. A new fallible `value_to_text_sink()` makes such a value go **loud** at string interpolation (`"x=$b"`), bare-word external argv (`prog $b`), and `echo` — instead of rendering the `[binary: N bytes]` placeholder where the real bytes should be. Valid-UTF-8 bytes still coerce; `==`/`in`/`${#…}`/`case` are unchanged. Remaining sinks (path-positionals, env, redirect target, semantic ops) are tracked for 0.11.1 (#93).

**interpreter — dead-code removal:** deleted the `Executor` trait, `NoOpExecutor`, `eval_command_subst`, and the lossy `result_to_value`. The sync `Evaluator` is only ever built with `NoOpExecutor`, and its live callers pre-resolve every operand to a literal via the async evaluator before sync evaluation — so the sync command-substitution / `eval_command` / FileTest arms were unreachable (and `result_to_value` was a latent lossy-`Bytes` trap). `Evaluator` is now non-generic; the three unreachable arms **fail loud on the outside case** rather than silently no-op'ing or stat'ing via `std::fs` (a VFS bypass). Drops the unused `Executor`/`NoOpExecutor` from the public `interpreter` surface (pre-1.0, breaking-in-name only — nothing implemented or consumed them).

**Round 1** (first commit) also fixed: backend `ToolResult`→`ExecResult` dropping `.data`/`content_type`/`baggage` (and masking a real backend error as exit-127); `source`/PATH `.kai` scripts keeping only the last statement's stdout; `Kernel::reset()` reverting latch/trash/`$$`/`initial_vars`; `jq` JSON-sniffing a `Value::String` `.data`; a non-ASCII short-flag argv bundle panicking mid-codepoint; a bareword `key=value` after `--` failing to parse; and `gather`'s own trailing redirect being dropped mid-pipeline.

## Notes

- **Rebased onto `main` (incl. #91 redirect-in-`$()` / clears-`.data`).** No textual conflicts; I re-verified the semantic seam by hand — #91's features and these guards coexist correctly.
- **Deferred to #93** (0.11.1): the binary-at-remaining-sinks cluster (a live but broader class), `jq -s` scalar-`.data` parity, `did_spill`/`original_code` across the `ToolResult` seam, and two LOW edges.
- Gates green: `cargo test --all` (0 failures), `cargo clippy --all --all-targets` (0 warnings), `cargo insta test --check`. Loud paths verified against the real binary.

A final kaibo review of the whole diff is running; its findings + any remediation will be posted as PR comments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -228,7 +228,7 @@ breaking entries are marked **BREAKING**.
   `Value::Bytes` value (e.g. `b=$(cat some-binary-file)`, since `cat` emits raw
   bytes for non-UTF-8 content) used to render the placeholder `[binary: N bytes]`
   when interpolated into a string (`"x=$b"`), passed as a bare word to an external
-  command (`prog $b`), or printed by `echo`. These text sinks now error (`binary
+  command (`prog $b`), or printed by `echo`/`printf`. These text sinks now error (`binary
   data (N bytes) cannot be used as text — decode it (base64/xxd) or redirect to a
   file`); valid-UTF-8 bytes still coerce. Semantic uses (`==`, `in`, `${#…}`,
   `case`) are unchanged. Path-positional/env/redirect sinks remain a known gap.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -224,6 +224,52 @@ breaking entries are marked **BREAKING**.
   demotion of `\|` to plain `|`).
 
 ### Fixed
+- **Binary data at text sinks is loud, not a silent placeholder** — a
+  `Value::Bytes` value (e.g. `b=$(cat some-binary-file)`, since `cat` emits raw
+  bytes for non-UTF-8 content) used to render the placeholder `[binary: N bytes]`
+  when interpolated into a string (`"x=$b"`), passed as a bare word to an external
+  command (`prog $b`), or printed by `echo`. These text sinks now error (`binary
+  data (N bytes) cannot be used as text — decode it (base64/xxd) or redirect to a
+  file`); valid-UTF-8 bytes still coerce. Semantic uses (`==`, `in`, `${#…}`,
+  `case`) are unchanged. Path-positional/env/redirect sinks remain a known gap.
+- **Binary worker output no longer corrupts `gather`** — scatter/gather result
+  rows and `gather --lines` lossily UTF-8-decoded (`U+FFFD`) a worker whose stdout
+  was binary; a row now reports `ok:false` with a clear error and `--lines`
+  hard-errors (exit 123) instead of emitting mangled bytes.
+- **`scatter`/`gather` reject wrong-typed flag values loudly** — `--limit`,
+  `--as`, and `--timeout` silently fell back to their defaults when handed a
+  value of the wrong type (a string `--limit "5"` from a variable, a negative
+  `--timeout`) or a `$(...)` the reduced sync binder can't evaluate; they now
+  coerce where sensible (`--limit "5"` → 5) and otherwise fail with a usage error
+  instead of running mis-configured.
+- **`scatter` workers inherit the parent execution context** — workers ran with a
+  from-scratch context that dropped the script watchdog (so a `ctx.patient` hold
+  suspended a missing clock → false-positive request-timeout kills) and bypassed
+  the output-limit memory cap (N parallel workers could each hold unbounded
+  output). Workers now derive from the parent, carrying the watchdog, output limit
+  (with a per-worker spill boundary), aliases, and ignore/external-command policy.
+- **Backend-tool results keep their structured data across the kernel seam** — a
+  backend-registered tool's `ToolResult` → `ExecResult` conversion carried only
+  `.out`, dropping `.data`/`content_type`/`baggage` (so `x=$(embedder_tool)` lost
+  a structured/collection result and OTel baggage). A symmetric conversion now
+  preserves them, and a real backend execution error is reported as a failure
+  rather than masked as exit-127 "command not found".
+- **`source` / PATH `.kai` scripts accumulate output across statements** — they
+  used to keep only the last statement's stdout, silently discarding everything
+  earlier.
+- **`Kernel::reset()` preserves embedder state** — it silently reverted the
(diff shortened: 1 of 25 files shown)


tobert:
## kaibo review (DeepSeek, holistic, whole changed surface)

**Verdict: conditional — one confirmed Medium gap (`printf`), one low test-only drift; everything else verified correct.**

### CONFIRMED — C1 (Medium): `printf` still renders binary as the placeholder
`crates/kaish-kernel/src/tools/builtin/printf.rs:34` — the `FormatArg for Value` impl's `as_format_string` renders `Value::Bytes` as `[binary: N bytes]`. `printf` is a pure text-output sink (like `echo`, which was fixed), so `printf "%s" $b` on a binary capture silently prints the placeholder instead of going loud. Reachable today. The infallible `fn as_format_string(&self) -> String` signature is the root cause — it forces the placeholder pattern.

### SUSPECTED — S1 (Low, test-only): `dispatch.rs` test twin drops non-literal positionals
`crates/kaish-kernel/src/dispatch.rs:216` — the test-only `build_args` twin handles only `Literal(String/Int/Float)` + `VarRef`, with `_ => {}` silently dropping `Bool`/`Json`/`Bytes`/`Interpolated`/`HereDocBody`. Production `build_args_flat` routes every positional through `value_to_text_sink`. A test passing here doesn't prove the real path — the exact "two spawn sites must stay in sync" gap.

(kaibo also raised an S2 on the sync `build_tool_args` WordAssign fallback and **retracted it itself** as currently-unreachable — noted for the 0.11.1 binary-sink cluster in #93.)

### Verified correct
- `value_to_text_sink` covers the five primary sinks (async + sync interpolation, external argv incl. the test twin, `echo`, pipeline-eval string parts).
- Scatter worker context: `child_for_pipeline()` + the three per-worker overrides (binding, `cancel = worker_token`, timer) — watchdog / output-limit / aliases / cancel-cascade all intact.
- Per-worker `spill_if_needed` boundary correct and non-leaking.
- Gather binary guard loud on both paths (JSONL row `ok:false`; `--lines` hard-error exit 123).
- Dead-code removal: all three now-loud sync arms (CommandSubst, Command, FileTest) confirmed unreachable in production — the async evaluator pre-resolves before sync eval.
- Wrong-typed scatter flags loud; `--progress` fully removed.

**Remediating C1 + S1 now** (both are the same loud-at-text-sink / sync-parity theme as this PR); pushing a follow-up commit.

tobert:
## Remediation of kaibo C1 + S1 — pushed `8be2985`

**C1 (Medium) — fixed.** `printf` is now a loud text sink: a binary (non-UTF-8) operand is rejected at the arg gate via `value_to_text_sink` (`printf: binary data (N bytes) cannot be used as text …`, exit 1) instead of formatting the `[binary: N bytes]` placeholder; valid-UTF-8 bytes coerce to their text. Kept the `FormatArg` trait infallible (it's shared with awk) and gated at ingestion, which is contained to `printf.rs`. Added a kernel-routed regression test `binary_arg_into_printf_is_loud`.

**S1 (Low, test-only) — fixed.** The `dispatch.rs` `BackendDispatcher` test twin's `_ => {}` now routes remaining literal positionals (`Bool`/`Json`/`Null`/`Bytes`) through `value_to_text_sink`, matching production `build_args_flat` — so a passing test proves the real path (the two-spawn-site sync discipline).

kaibo's retracted S2 and the broader binary-at-remaining-sinks class are tracked in #93 (0.11.1).

Gates green: `cargo test --all` 0 failures, `clippy --all --all-targets` 0 warnings, and verified against the real binary (`printf "%s" \$binary` → loud exit 1; normal `printf` unaffected).

PR #106, opened 2026-07-05
fix(grep): `-r` searches a file operand instead of silently missing (GH #105)

### LINKED ISSUE #105: grep -r PATTERN FILE (a file operand, not a dir) silently finds nothing — search the file directly
### What happens

`grep -r PATTERN FILE`, where the operand is a single **file** rather than a directory, returns **zero matches, exit 1, and no error** — even when the file plainly contains the pattern. Plain `grep PATTERN FILE` finds it.

Live on 0.11.0:

```
grep -rc 'kaish' AGENTS.md   # → 0    (exit 1, silent)
grep -c  'kaish' AGENTS.md   # → 15
grep -rc 'kaish' .           # → 662  (a directory works fine)
```

`-R` behaves the same (it aliases `-r`), and a `./` prefix doesn't change anything — the operand is still a file.

### Why it's worth fixing

`grep -r PATTERN somefile` is the form people and agents reach for *without thinking* — `-r` is muscle memory, and handing a file to a search-with-`-r` and having it search that file is the reflex the whole ecosystem shares. Nobody expects `-r` to mean "silently ignore a file argument."

The failure mode is the bad kind: it's **silent**. A caller — especially a model driving the shell — reads the empty result as "the pattern isn't there," a false negative that quietly corrupts whatever reasoning follows, with no error to catch it. That's much worse than a loud "not a directory."

Concrete downstream cost: kaibo carries a hand-rolled note in its shell preamble *purely* to steer models around this (`-r wants a directory; for one file use plain grep`). Fixing it here lets that note — and the instruction tokens it spends every session — go away.

### Root cause

In the recursive branch (`crates/kaish-kernel/src/tools/builtin/grep.rs`, ~L380–434), `-r` **unconditionally** treats the path operand as a walk root and enumerates it with a `**/*`, files-only `FileWalker`. A directory yields its files; a plain file has nothing "under" it, so the walk collects zero entries → `grep_multiple_files` over an empty list → 0 matches, exit 1. (`-r`/`-R` fold together at ~L253.)

### Proposed fix

Before building the walker, if the resolved operand is a file (`is_file(root)`), route it through the existing single-/multi-file search path — the non-recursive branch just below already does exactly this for explicit file operands — instead of the directory walker. `-r` then means "recurse into directories, search files directly": recursion only governs how *directories* expand, which is the behavior everyone already assumes.

Test to add in `grep_search_features_tests.rs`: `grep -r PATTERN <file-with-a-match>` returns the match (and cover `-R` too, plus a mixed `grep -r PATTERN file dir` operand list).

tobert:
Closes #105.

## The bug

`grep -r PATTERN FILE`, where the operand is a **file** rather than a directory, returned **zero matches, exit 1, and no error** — even when the file plainly contained the pattern. Live on 0.11.0:

```
grep -rc 'kaish' AGENTS.md   # → 0    (exit 1, silent)
grep -c  'kaish' AGENTS.md   # → 15
grep -rc 'kaish' .           # → 662  (a directory works fine)
```

The recursive branch (`grep.rs`) unconditionally treated the path operand as a walk root and enumerated it with a files-only `**/*`. A directory yields its files; a plain file has nothing "under" it, so the walk collected zero entries → `grep_multiple_files` over an empty list → 0 matches.

The failure mode is the bad kind — **silent**. A caller, especially a model driving the shell, reads the empty result as "the pattern isn't there": a false negative that quietly corrupts whatever reasoning follows. kaibo carries a hand-rolled shell-preamble note *purely* to steer models around this; fixing it here lets that note (and its per-session instruction tokens) go away.

## The fix

Reframe what `-r` means: it governs how **directories** expand, nothing more — "recurse into directories, search files directly," which is the behavior everyone already assumes. The recursive branch now partitions its operands by `stat`:

- **file** → searched directly (falls through to the ordinary file-operand path below, so a lone file prints unprefixed like plain `grep -c`);
- **directory** → walked;
- **mixed** `grep -r p file dir` → does both (file operand joined to the walked set);
- **stat-unresolvable** operand → still falls to the walker, preserving pre-fix behavior for a bad root.

Display is deliberately conservative: the sole-directory walk strips its root **byte-for-byte as before** (zero change to the overwhelmingly common `grep -r p dir` output); only the genuinely-new mixed/multi-source shapes switch to cwd-relative prefixes so each source shows under its own subpath (`top.txt`, `sub/inner.txt`).

## Tests

Four kernel-routed regression tests in `grep_search_features_tests.rs`, written TDD-first (red → green): single file operand, `-rc` count (the exact `→ 0` repro), the `-R` alias, and a mixed `file dir` operand list. Also verified end-to-end against the built binary.

Full suite (4370 tests) + `cargo clippy --all --all-targets` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,15 @@ breaking entries are marked **BREAKING**.
 
 ## [Unreleased]
 
+### Fixed
+- **`grep -r PATTERN FILE`** (a file operand, not a directory) now searches
+  that file instead of silently finding nothing (GH #105). `-r`/`-R` used to
+  treat every operand as a walk root; a file has nothing "under" it, so the
+  walk collected zero entries → 0 matches, exit 1, no error — a false negative
+  a model reads as "not found". `-r` now governs only how *directories* expand:
+  files are searched directly, directories walked, and a mixed `grep -r p file
+  dir` operand list does both.
+
 ## [0.11.0] - 2026-07-04
 
 ### Added
--- a/crates/kaish-kernel/src/tools/builtin/grep.rs
+++ b/crates/kaish-kernel/src/tools/builtin/grep.rs
@@ -379,59 +379,107 @@ impl Tool for Grep {
 
         // Handle recursive search
         if recursive {
-            let path = args
-                .get_string("path", 1)
-                .unwrap_or_else(|| ".".to_string());
-            let root = ctx.resolve_path(&path);
-
-            // Build include/exclude filter
-            let mut filter = IncludeExclude::new();
-            if let Some(Value::String(inc)) = args.get("include", usize::MAX) {
-                filter.include(inc);
-            }
-            if let Some(Value::String(exc)) = args.get("exclude", usize::MAX) {
-                filter.exclude(exc);
-            }
-
-            // Build glob pattern if include is specified
-            let glob = if let Some(Value::String(inc)) = args.get("include", usize::MAX) {
-                GlobPath::new(&format!("**/{}", inc)).ok()
+            // Partition the operands by kind. A *directory* is walked; a *file*
+            // is searched directly (#105 — a file has nothing "under" it, so
+            // treating it as a walk root collected zero entries and reported no
+            // matches *silently*, the worst failure mode for a model reading the
+            // empty result as "not found"). `-r` governs how *directories*
+            // expand, which is the only thing anyone means by it. A missing or
+            // otherwise-unstattable operand falls to the walker, preserving the
+            // pre-#105 behavior for a bad root.
+            let operands: Vec<String> = args
+                .positional
+                .iter()
+                .skip(1)
+                .map(crate::interpreter::value_to_string)
+                .collect();
+            let operands = if operands.is_empty() {
+                vec![".".to_string()]
             } else {
-                GlobPath::new("**/*").ok()
+                operands
             };
 
-            let options = WalkOptions {
-                max_depth: None,
-                entry_types: crate::walker::EntryTypes::files_only(),
-                respect_gitignore: ctx.ignore_config.auto_gitignore(),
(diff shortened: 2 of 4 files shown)


tobert:

Verdict: **production-quality.** Confirmed against current code:

- **Fall-through control flow correct** — file-only operands under `-r` reach the ordinary non-recursive paths (single file → unprefixed, multiple → prefixed); recursive-only flags (`--include`/`--exclude`/`--ftype`/`--hidden`) are correctly inert there, and `--ftype` names are still validated regardless.
- **Borrow/lifetime sound under NLL** — the `ctx.backend` borrow via `BackendWalkerFs` ends at its last loop use, so the `&mut ctx` reborrow at `grep_multiple_files` is fine.
- **Symlinks / glob expansion / display prefix** — all cases correct (file symlink searched directly, dir symlink walked, broken symlink → walker; mixed shapes cwd-relative).
- **Duplication** (operand derivation at two sites) — trivial, no existing helper to reuse; extracting a `partition_operands` helper would be premature (grep is the only caller).

### Known limitation (recorded, not fixed — 80% rule)

`report_missing` asymmetry for an **explicit file operand**: `grep -r p file1 file2` reports an unreadable/typo'd operand loudly (exit 2), but `grep -r p file dir/` passes `report_missing=false` for the whole combined set, so an unreadable explicit `file` is silently skipped (or, if it never existed, it falls to the walker → exit 1 with a walker error rather than a clear "No such file"). Defensible: in the mixed path the file operands were confirmed to exist by `stat`, so a later read failure is a genuine TOCTOU race; and this matches the pre-#105 behavior for a bad root. The strict fix (per-file `report_missing`, i.e. two `grep_multiple_files` calls whose results must be merged) costs more than the rare race warrants. Left as-is; flagging for the record.

PR #107, opened 2026-07-05
fix(redirects): run `$()` targets on a bare `Kernel::execute` — thread the dispatcher (GH #90)

### LINKED ISSUE #90: `$(...)` in a redirect target fails for a bare single command
While adding redirect-inside-`$()` support (redirect-in-cmdsubst branch), I found a pre-existing bug: command substitution in a redirect **target** silently fails when the command is a *bare single command* (not part of a pipeline or `&&`/`||` chain).

## Reproduce

```
echo x > $(echo f.txt)
# → exit 1: "redirect: could not evaluate redirect target"; f.txt is NOT created
```

## Works (dispatcher attached)

```
echo x | cat > $(echo g.txt)          # pipeline → g.txt created
echo x > $(echo /tmp/gen) && cat ...  # && chain → works (see sandbox_mode_tests.rs:154)
```

## Root cause

`eval_redirect_target` (`crates/kaish-kernel/src/scheduler/pipeline.rs:170-176`) runs `$()` in the target only when `ctx.dispatcher` is `Some`; otherwise it falls back to `eval_simple_expr`, which cannot evaluate command substitution and returns the "could not evaluate redirect target" error. For a bare single-command statement the dispatcher is not attached, so the fallback fires.

## Impact

`docs/LANGUAGE.md` documents `cmd > $(gen-path)` as working. It does — but only inside a pipeline or chain; the standalone form fails loudly. It is also the reason a nested `$(cmd > $(subst))` cannot execute (the parser handles the nesting fine; only the target-eval step fails).

## Fix direction

Attach the dispatcher for single-command statements too (or route them through the runner that already attaches it), so redirect-target `$()` runs consistently regardless of surrounding syntax. Add a bare-command regression test (`echo x > $(echo f) ; cat f`).

tobert:
Closes #90.

## The bug — and why it hid

`$(...)` in a redirect **target** or heredoc body failed with `redirect: could not evaluate redirect target`:

```
echo x > $(echo f.txt)     # exit 1, f.txt NOT created
cat  < $(echo f.txt)       # same
echo x | cat > $(echo g)   # pipeline stage: same
```

…but **only through `kernel.execute()` on a `Kernel` held by value** — the path every embedder (kaibo/kaijutsu) and the entire test harness use. Through the REPL it worked, which is exactly why it looked already-fixed and the GH #90 comment in `redirect_in_cmdsubst_tests.rs` had gone stale.

`eval_redirect_target` read `ctx.dispatcher` to run the `$()`, and **silently fell back** to the sync evaluator (which can't run command substitution) when it was `None`. `ctx.dispatcher` is populated from `self.dispatcher()`, which upgrades a `self_weak` that is set **only in `into_arc`**. The REPL Arc-attaches its kernel → dispatcher present. A bare `Kernel::execute` never sets `self_weak` → dispatcher `None` → the target silently degraded to a wrong answer.

The pinning test written through `kernel.execute()` is what caught it — red on all three cases.

## The fix

Without an `Arc<Kernel>` there is no `Arc<dyn CommandDispatcher>` to put in `ctx.dispatcher`, so the owned-handle route fundamentally can't serve a bare kernel. But the **runner always holds a real `&dyn CommandDispatcher`** (the kernel itself, Arc'd or not). So thread it: through `apply_redirects` and `setup_stdin_redirects` into `eval_redirect_target`, which now calls `dispatcher.eval_expr` directly.

- The `ctx.dispatcher`-or-sync-fallback branch is **deleted** — no silent fallback (a missing dispatcher can no longer become a wrong answer; there's always a real one).
- A **pipeline-stage** target had the same gap (the stage ctx copies `dispatcher: None`); fixed by passing the stage's own forked dispatcher.
- `ctx.dispatcher` the field stays (the `timeout` builtin's re-dispatch still uses it); only the redirect evaluator stops depending on it.

Behavior no longer depends on how the kernel was constructed.

## Tests

Four kernel-routed regression tests in `redirect_in_cmdsubst_tests.rs`, TDD-first (red → green): bare-command target, `<` stdin target, the nested `$(cmd > $(subst))` form #90 called out, and a pipeline-stage target. The stale "GH #90 blocked" comment is replaced with the real pins.

Full suite (**4374**) + `cargo clippy --all --all-targets` clean. REPL smoke-tested unregressed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,15 @@ breaking entries are marked **BREAKING**.
 ## [Unreleased]
 
 ### Fixed
+- **`$(...)` in a redirect target now works on a bare `Kernel::execute`** (GH
+  #90). Command substitution in a redirect target or heredoc body (`echo x >
+  $(gen)`, `cat < $(gen)`, `cmd > $(gen)` in a pipeline stage) only ran when the
+  kernel was Arc-attached via `into_arc` — the REPL. A bare `Kernel::execute`
+  (every embedder holding a `Kernel` by value, and the whole test harness) left
+  `ctx.dispatcher` unset, so the target silently fell back to a sync evaluator
+  that can't run `$()` and failed with "could not evaluate redirect target".
+  The redirect evaluator now takes the dispatcher the runner already holds, so
+  the behavior no longer depends on how the kernel was constructed.
 - **`grep -r PATTERN FILE`** (a file operand, not a directory) now searches
   that file instead of silently finding nothing (GH #105). `-r`/`-R` used to
   treat every operand as a walk root; a file has nothing "under" it, so the
--- a/crates/kaish-kernel/src/scheduler/pipeline.rs
+++ b/crates/kaish-kernel/src/scheduler/pipeline.rs
@@ -30,6 +30,7 @@ pub(crate) async fn apply_redirects(
     mut result: ExecResult,
     redirects: &[Redirect],
     ctx: &ExecContext,
+    dispatcher: &dyn CommandDispatcher,
 ) -> ExecResult {
     // Defer materialization of OutputData → result.out to individual redirect
     // handlers. File redirects (Overwrite/Append) can stream OutputData directly
@@ -71,7 +72,7 @@ pub(crate) async fn apply_redirects(
                 result.clear_stdout();
             }
             RedirectKind::StdoutOverwrite => {
-                let path = match eval_redirect_target(&redir.target, ctx).await {
+                let path = match eval_redirect_target(&redir.target, ctx, dispatcher).await {
                     Ok(p) => p,
                     Err(e) => return ExecResult::failure(1, format!("redirect: {e}")),
                 };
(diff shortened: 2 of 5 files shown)


tobert:

Verdict: **correct, safe, and complete.** Verified against current code:

- **Forked `task_dispatcher` for a pipeline-stage `apply_redirects` is correct — and *more* consistent than the old code.** `dispatch_command` syncs `ctx.scope → self.scope` before running and back after, so by the time `eval_expr` runs the forked kernel's scope reflects the stage's ctx. The old `stage_ctx.dispatcher` (cloned from the parent) meant an *intermediate* stage's `$()` target read the parent's not-yet-updated scope; the new path is self-consistent within the stage. `$()` snapshot/restore (`kernel.rs:3918`) contains any `cd`/assignment inside the target.
- **Deleting the `eval_simple_expr` fallback is strictly better** — it replaces a silent degradation on bare kernels with "works (Kernel dispatcher) or fails loud (test-only `BackendDispatcher`, documented limitation)". No test passes a `$()` target through `BackendDispatcher`.
- **`Kernel::eval_expr` scope/cwd correct for all three paths** (bare command / pipeline stage / scatter-gather).
- **No borrow/`Send`/`'static` issue** with `&*task_dispatcher` inside `tokio::spawn(async move …)` — the `Arc` is moved in; the `&dyn` doesn't escape the `.await`.
- **Complete inventory of remaining `ctx.dispatcher` readers**: `timeout.rs` (re-dispatch — separate concern, and it fails *loud* on a bare kernel) and the `dispatch_command`/background/fork *writes*. No other redirect/heredoc path depends on `ctx.dispatcher`; the fix is complete.

No action items.

Issue #92, opened 2026-07-04
Plumb the confirmation latch to a first-class typed public-API field (off `.data`)


tobert:
## Problem

The confirmation-latch request (`set -o latch` → `rm` exit-2 + nonce) rides inside `ExecResult.data` as serialized JSON, decoded back via `ExecResult::latch_request()`. `.data` is thus **overloaded**: for `seq`/`jq`/`fromjson` it's the structured view of stdout (data-plane); for a latched `rm` it's a confirmation nonce (control-plane).

This overloading forced a workaround in #91 (redirects inside `$()`): a stdout redirect must clear the data-plane `.data`, but blanket-clearing it silently disabled the safety gate on `rm precious > log`. #91 papers over it with a discriminator in `ExecResult::clear_stdout()`:

```rust
if self.latch_request().is_none() {
    self.data = None;
}
```

Correct, but it's a runtime shape-check standing in for a structural boundary. Two important features (redirect semantics + the safety latch) share one field and weaken each other's boundaries.

## Proposal

Give the latch its own typed field so the planes are structural, not shape-sniffed:

- Add `latch: Option<LatchRequest>` to `ExecResult` **and** `ToolResult` (it must survive the `ExecResult`↔`ToolResult` backend roundtrip — see backend.rs `From<ExecResult> for ToolResult` and the dispatch reconstruction; cf. #84 which already tracks that boundary dropping fields).
- `ExecContext::latch_result()` sets `.latch` (typed) instead of serializing into `.data`.
- `ExecResult::latch_request()` becomes `self.latch.clone()`.
- `clear_stdout()` drops the discriminator — `.data` becomes unambiguously data-plane and clears unconditionally.

Net win on two axes: (1) plane separation; (2) the latch stops being round-tripped JSON and becomes typed end-to-end.

## Costs / breaking

- **BREAKING (embedder / `--json`):** a latched result's serialized shape changes — the nonce moves off `.data` (no longer nested under the `--json` error envelope). Embedders using the sanctioned `latch_request()` accessor are insulated; no test currently pins the `--json` latch envelope. Needs a `**BREAKING:**` CHANGELOG entry.
- ~20 mechanical constructor edits across `ExecResult`/`ToolResult` (both derive/spell all fields).
- **Docs:** rework the latch/embedding docs (EMBEDDING.md, LANGUAGE.md latch section, `LatchRequest` rustdoc) to describe the typed field as the contract.

## Follow-up hook

[...]

PR #108, opened 2026-07-05
feat(#96): surface the backgrounded confirmation latch to its consumers

### LINKED ISSUE #96: Background jobs store the confirmation latch but no consumer surfaces it
Found during the latch-first-class review (GH #92, cross-model). The `.latch` field propagates correctly *into* background-job storage, but no consumer exposes it, so a backgrounded gated op can't be confirmed.

## Repro

```
set -o latch
rm precious.txt &      # job gates (exit 2 + latch) — stored, never surfaced
wait                   # returns non-zero, but the nonce is unreachable
```

The full `ExecResult` (including `.latch`) is delivered over the oneshot and held in `Job::result`, but:

- **`wait`** (`tools/builtin/wait.rs`) only reads `result.ok()` / the exit code — never `latch_request()`.
- **`jobs` / `JobInfo`** (`kaish-types/src/job.rs`) has no `latch` field.
- **`/v/jobs/{id}/`** (`vfs/jobfs.rs`) has no `latch` node.
- **`Job::status()`** maps exit 2 to `JobStatus::Failed`, indistinguishable from any other failure.

## Impact

Low/safe: the destructive op stays **blocked** (the latch did its job), and `wait` returns non-zero — no data loss, no silent bypass. The limitation is that a *backgrounded* latch can't be **fulfilled** (`Kernel::confirm`) because the request never reaches a consumer. Backgrounding a gated destructive op under `set -o latch` is an unusual pattern.

## Fix direction

Surface the stored latch on the job consumers: `wait` returns/exposes the `LatchRequest` (or its exit-2 result verbatim), `JobInfo` gains a `latch: Option<LatchRequest>`, and `/v/jobs/{id}/latch` renders it. Then an embedder can `Kernel::confirm` a backgrounded gate. Distinguish `JobStatus` exit-2-latch from a plain failure.

tobert:
Closes #96.

## The gap

The confirmation latch went first-class in #92, but a backgrounded gate was a dead end. `rm precious.txt &` under `set -o latch` gates in the background and **stores** its `LatchRequest` in the job's result — but every consumer was blind to it:

- **`wait`** reported `Failed` (read only the exit code).
- **`jobs` / `JobInfo`** had no latch field.
- **`/v/jobs/{id}/`** had no latch node.
- **`Job::status()`** folded exit-2-latch into `Failed`, indistinguishable from a real error.

The op stayed safely **blocked** (no data loss), but the nonce was unreachable, so a backgrounded gate could never be **fulfilled** (`Kernel::confirm`).

## The fix — thread the stored latch to all four surfaces

| Surface | Change |
|---|---|
| `JobStatus` | new `Latched` variant — the state is *held*, not errored |
| `jobs` / `/v/jobs/{id}/status` | report `Latched` / `"latched"` distinctly |
| `JobInfo` | new `latch: Option<LatchRequest>` — the programmatic surface (`JobManager::list`/`get`) |
| `wait` | surfaces the request on the result's control-plane `.latch` field, exit 2 — mirrors a foreground gate so `latch_request()` / `Kernel::confirm` work identically |
| `/v/jobs/{id}/latch` | **new** VFS node: the request as JSON (nonce, command, paths, hint); empty body when not gated |

Design touches: `wait` on several gated jobs keeps the *first* latch (`.latch` holds one; waiting on multiple gated jobs is an odd pattern); the VFS node reads empty (not error) when not latched, matching how `status`/`command` always read.

## The capstone test

Drives the whole loop end to end: background a gated `rm`, `wait 1` to surface the latch, `Kernel::confirm(&latch)`, assert the file is gone — from backgrounded gate to fulfillment in one test. Plus surface tests for `jobs`/status/the JSON node.

Docs updated (help `vfs.md`, `EMBEDDING.md`, `LANGUAGE.md`). 4377 suite + `cargo clippy --all --all-targets` clean; verified end-to-end through the REPL (`wait`→Latched, status→"latched", latch node → full JSON).

**BREAKING (embedders):** `JobStatus` gains a `Latched` variant (exhaustive matches must handle it) and `JobInfo` gains a `latch` field.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,19 @@ breaking entries are marked **BREAKING**.
 
 ## [Unreleased]
 
+### Added
+- **A backgrounded confirmation latch is now surfaced and fulfillable** (GH
+  #96). A destructive op gated under `set -o latch` and run in the background
+  (`rm x &`) stored its `LatchRequest` but no consumer exposed it, so the nonce
+  was unreachable and the gate could never be confirmed. Now: `wait` surfaces
+  the request on the result's `.latch` field (exit 2, like a foreground gate);
+  `jobs` and `/v/jobs/{id}/status` report `Latched` distinctly from a plain
+  failure; and a new `/v/jobs/{id}/latch` node renders the request as JSON
+  (nonce, command, paths, hint) — empty when the job isn't gated. An embedder
+  fulfills it with `Kernel::confirm(&latch)`. **BREAKING (embedders):**
+  `JobStatus` gains a `Latched` variant (exhaustive matches must handle it) and
+  `JobInfo` gains a `latch: Option<LatchRequest>` field.
+
 ### Fixed
 - **`$(...)` in a redirect target now works on a bare `Kernel::execute`** (GH
   #90). Command substitution in a redirect target or heredoc body (`echo x >
--- a/crates/kaish-help/content/en/vfs.md
+++ b/crates/kaish-help/content/en/vfs.md
@@ -33,9 +33,15 @@ Each background job gets a directory:
 ```
 /v/jobs/{id}/stdout    live output (ring buffer, 10MB max)
 /v/jobs/{id}/stderr    live error stream
-/v/jobs/{id}/status    "running" | "done:0" | "failed:N"
+/v/jobs/{id}/status    "running" | "done:0" | "latched" | "failed:N"
 /v/jobs/{id}/command   original command string
+/v/jobs/{id}/latch     confirmation-latch request (JSON) if gated, else empty
 ```
+
+A destructive op backgrounded under `set -o latch` (`rm x &`) gates in the
+background: status is `latched`, and `/v/jobs/{id}/latch` carries the JSON
+request (nonce, command, paths) so the gate can be fulfilled — read the nonce
+and re-run with `--confirm=<nonce>`, or, from an embedder, `Kernel::confirm`.
 
 ```sh
 cargo build &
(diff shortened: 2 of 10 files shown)


tobert:

Verdict: **no gap, no misclassification, no redundant work, no missed consumer.** Verified against current code:

- **`JobStatus::Latched` classification correct & consistent** — `code` and `latch` are orthogonal `ExecResult` fields; the only producer (`ToolCtx::latch_result`) always pairs `code=2` + `latch=Some`, and a plain exit-2 usage error (no latch) correctly falls to `Failed`. `status()` and `status_string()` triage identically.
- **`JobInfo` construction sound** — `try_poll()`'s early-out (`if self.result.is_some()`) means `job.latch()` after `job.status()` is a no-op second poll; no lock/borrow concern (closure is sync, no `await` under the mutex).
- **`wait` mirrors a foreground gate** — `finish()` produces the same `code:2 + latch:Some(Box)` shape as `ToolCtx::latch_result`, so `latch_request()`/`Kernel::confirm` behave identically. "First latch wins" is a defensible single-`.latch` tradeoff (the capstone test pins the full loop).
- **VFS empty-body-when-not-latched is the right call** — `stat` reports size 0 for all dynamic nodes anyway, so a consumer must read; empty-vs-JSON cleanly separates "not gated" from "gated", and `NotFound` still means "no such job".
- **Complete consumer audit** — `jobs` renders `Latched` via `Display`; `kill`/`fg`/`bg` never inspect `JobStatus` (correct); `running_count` excludes latched (`!is_done`); the only exhaustive `match` is the `Display` impl (all 5 variants handled).

### Noted edge (acceptable, for the record)
`jobs --cleanup` / `JobManager::cleanup()` treats a latched job as `is_done()` and removes it — discarding an unread latch. Acceptable: cleanup is opt-in, and the latch is surfaced first via `wait`/`wait_all` (shutdown uses `wait_all`, not `cleanup`) and the VFS node. Not fixed.

No action items.

Issue #48, opened 2026-06-30
Interpreter memory/allocation optimization pass (tokio + ExecContext)


tobert:
## Abstract

A general memory/allocation optimization pass on the hot execution path. We suspect
steady-state cost is low with occasional peaks, so this is not urgent — but we've never
done a deliberate tokio + allocation sweep, and the per-command dispatch chain carries
some obvious avoidable weight. File this as the home for "make the interpreter cheaper per
command" work: profile first, then trim.

## Known weight (starting points, not a mandate)

- **`ExecContext` is large and copied repeatedly.** `execute_pipeline` builds a ~30-field
  struct on the stack (`kernel.rs:2591`); `dispatch_command` copies ~25 fields field-by-field
  **twice** — in (`:5089`) and out (`:5134`) — per command, paid at every recursion level
  above it. Candidate for sharing/borrowing instead of cloning.
- **`#[tracing::instrument]` / `debug_span` future bloat** on hot frames
  (`execute_command_depth` `:2829`, `execute_pipeline` `:2566`, `execute_stmt_flow`'s
  manual span `:2050`, and `pipeline.rs:291/357/397`). Instrumenting an `async fn` enlarges
  its generated future with span state. Worth measuring whether to demote some to coarser
  spans on the per-command path.
- **Boxed async state machines.** Every recursive `eval_*_async` / `execute_stmt_flow`
  level is a `Box::pin` heap allocation holding all live locals across `.await`s. Mostly
  structural (Rust async recursion requires boxing), but the frame *size* is reducible by
  shrinking what each future holds.

## Approach

Profile a representative workload (heap + allocation count) to find the real peaks before
changing anything, then do basic tokio/allocation passes against whatever the profile
flags. Treat the bullets above as hypotheses to confirm, not a task list.

Related: #46 (depth guard), #47 (worker-thread stack size) — those are correctness/
robustness; this is pure efficiency.

tobert:
## Per-level recursion stack cost — batch review findings (2026-07-05)

[...]

PR #109, opened 2026-07-05
feat(#46/#47): depth-guard the statement engine + size execution threads for it

### LINKED ISSUE #46: Guard statement-engine recursion against stack overflow
## Abstract

The async statement engine recurses without any depth guard. Deeply nested `$(...)`
command substitution, recursive shell functions, and recursive `.kai` scripts all
re-enter the interpreter on the same stack and will overflow it — a `SIGSEGV` with no
diagnostic, rather than a clean error. We already guard analogous recursion elsewhere
(alias re-entry caps at 10, the lexer caps `$((...))` nesting at `MAX_PAREN_DEPTH = 256`);
the statement engine has nothing equivalent.

## Why it matters

Crashing is acceptable; a silent segfault with no message is not. A pathological or
buggy script (or an agent generating one) should hit a loud, catchable "max recursion
depth exceeded" error, not take the process down with an opaque signal.

## Where the recursion is unbounded

All in `crates/kaish-kernel/src/kernel.rs`:

- **Command substitution** — `Expr::CommandSubst(Vec<Stmt>)` (`:3670`, also string-interp
  `StringPart::CommandSubst` `:3934`) re-enters `execute_block_capturing` (`:4129`) →
  `execute_stmt_flow` (`:2045`). Two boxed async futures stack per nesting level.
- **Shell functions** — `execute_user_tool` (`:4014`) loops `execute_stmt_flow` (`:4060`)
  in the same kernel (no fork / new task).
- **`.kai` scripts** — `try_execute_script` (`:4287`) loops `execute_stmt_flow` (`:4380`);
  a `.kai` calling a `.kai` recurses unbounded.

## Existing guards to mirror

- Alias re-entry: `if alias_depth < 10` in `execute_command_depth` (`kernel.rs:2840`).
- Lexer: `MAX_PAREN_DEPTH = 256` (`lexer.rs:23`), explicitly "Prevents stack overflow
  from pathologically nested inputs."

## Sketch

Thread a depth counter (or reuse/extend the per-execution context) through
`execute_stmt_flow` / command-substitution / function-call / script-source entry, and
return a loud error past a cap. The validator already tracks `loop_depth`/`function_depth`
(`validator/walker.rs:25`) for semantic checks — worth deciding whether the cap belongs
there (static, pre-execution) or in the engine (dynamic, covers runtime recursion like
mutual `$()`/function recursion). Related: #47 (worker-thread stack size).

### LINKED ISSUE #47: Set an explicit tokio worker-thread stack size
## Abstract

The tokio runtime is built with `Runtime::new()` defaults and no explicit worker-thread
stack size. The interpreter's deep recursion (command substitution, functions, `.kai`
scripts) runs on whatever stack the executing thread has. The foreground root future runs
on the `block_on` thread, but anything that hops onto a spawned task — piped pipeline
stages, scatter/gather workers, background jobs — runs on a tokio **worker thread with the
default ~2 MB stack**. So the same script has materially less headroom inside a pipe or a
background job than in the foreground, with no obvious reason from the user's side.

## Why it matters

Stack capacity should be intentional and uniform, not an accident of whether code happens
to run on the `block_on` thread vs. a spawned task. Pairs with #46 (depth guard): a
guard sets the policy, an explicit stack size sets the budget.

## Where

- `crates/kaish-repl/src/main.rs:227`, `:257` — `tokio::runtime::Runtime::new()?`
- `crates/kaish-repl/src/lib.rs:514`, `:531` — `Runtime::new()`
- Forked work lands on worker threads via spawned tasks: pipeline stages
  (`scheduler/pipeline.rs:427`), scatter/gather (`pipeline.rs:312`), background jobs.

## Sketch

Build the runtime with `runtime::Builder::new_multi_thread().thread_stack_size(...)` (and
consider matching `worker_threads` to an embeddable default) so foreground and worker
stacks are the same, chosen value. Decide the number deliberately rather than inheriting
2 MB. Keep it in sync across the `main.rs` and `lib.rs` construction sites.

tobert:
Closes #46. Closes #47.

## #46 — the bug

The async statement engine recursed with **no bound**. Deeply nested `$(...)`, recursive shell functions, and `.kai` scripts sourcing each other re-enter the interpreter on the native stack and overflow it — a bare `SIGSEGV`/abort with no diagnostic:

```
$ kaish -c 'f() { f; }; f'
thread 'main' has overflowed its stack
fatal runtime error: stack overflow, aborting   # exit 134
```

We guard the analogous cases elsewhere (alias re-entry caps at 10, the lexer caps `$((…))` nesting at 256); the statement engine had nothing.

## The guard (#46)

Small and mechanical: an `AtomicUsize` on the Kernel (fresh per fork, so each chain is bounded independently), an RAII decrement so a cancelled/errored re-entry stays balanced, and an `enter_recursion()` check at the three re-entry points — `execute_block_capturing` (`$()`), `execute_user_tool` (functions), `try_execute_script` (`.kai`). Past the cap it returns a **loud, catchable** error:

```
$ kaish -c 'f() { f; }; f'
maximum recursion depth (32) exceeded in a shell function — …   # exit 1
```

## Why #47 is in the same PR

Picking the cap forced it. We measured native stack cost per level (disable the cap, probe depth-at-overflow):

| thread | debug (~380 KB/level) | release (~80 KB/level) |
|---|---|---|
| 8 MB main | ~21 levels | ~100 |
| 2 MB tokio worker | ~5 | ~25 |

On the default stacks, **no single cap both protects a 2 MB worker and allows useful depth** — and a `#[tokio::test]` thread (~2 MB) would SIGSEGV before a recursion test could even assert the guard. The guard needs a controlled stack to be effective **and** testable.

## The floor (#47)

The cap (`MAX_RECURSION_DEPTH` = 32) is tuned for a documented floor (`RECOMMENDED_STACK_SIZE` = 16 MiB), both now `pub` for embedders. kaish can't set stack size itself (it doesn't own the runtime), so the reference REPL does:
- `thread_stack_size(RECOMMENDED_STACK_SIZE)` on its tokio worker threads (pipeline/scatter/bg recursion);
- a `std::thread` with the recommended stack driving `block_on` (the OS main thread's ~8 MB overflows a deep debug recursion before the cap).

`docs/EMBEDDING.md` tells embedders to size their execution threads the same way (with a code sketch); `limits.md` documents the cap.

## Tests

Each recursion runs on a `RECOMMENDED_STACK_SIZE` `std::thread` with a fresh current-thread runtime — the only way to reach the cap without the overflow-under-test aborting the binary. Runaway self-recursion and mutual recursion go loud; depth-16 bounded recursion succeeds; the counter is balanced across executes (no leak). Verified end-to-end in **debug and release**, foreground and inside a pipeline stage. 4378 suite + `cargo clippy --all --all-targets` clean.

## Consistency note

A recursion error *inside* `$(...)` follows kaish's existing command-substitution semantics — a failed `$()` yields an empty expansion and doesn't fail the enclosing command (same as `echo $(false)`), so that path is **bounded** (no crash) but not exit-code-loud; direct recursion (functions/scripts) is exit-1-loud. Not a new silent path.

## Follow-up

Reduce per-level stack cost (#48) so the floor can drop — a gemini-pro + fable batch review of the recursion path is in flight.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,8 +23,9 @@ breaking entries are marked **BREAKING**.
   `JobStatus` gains a `Latched` variant (exhaustive matches must handle it) and
   `JobInfo` gains a `latch: Option<LatchRequest>` field.
 - **Recursion is depth-guarded (`MAX_RECURSION_DEPTH` = 32)** (GH #46/#47).
-  Command substitution, shell-function calls, and `.kai` script sourcing all
-  re-enter the statement engine on the native stack; a runaway or mutually
+  Command substitution, shell-function calls, `.kai` script execution, and
+  `source`/`.` all re-enter the statement engine on the native stack; a runaway
+  or mutually
   recursive script now returns a loud `maximum recursion depth exceeded` error
   instead of overflowing the stack (a `SIGSEGV`/abort with no diagnostic). Two
   new `pub const`s let embedders size their runtime: `MAX_RECURSION_DEPTH` and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,8 +22,21 @@ breaking entries are marked **BREAKING**.
   fulfills it with `Kernel::confirm(&latch)`. **BREAKING (embedders):**
   `JobStatus` gains a `Latched` variant (exhaustive matches must handle it) and
   `JobInfo` gains a `latch: Option<LatchRequest>` field.
+- **Recursion is depth-guarded (`MAX_RECURSION_DEPTH` = 32)** (GH #46/#47).
+  Command substitution, shell-function calls, and `.kai` script sourcing all
+  re-enter the statement engine on the native stack; a runaway or mutually
+  recursive script now returns a loud `maximum recursion depth exceeded` error
+  instead of overflowing the stack (a `SIGSEGV`/abort with no diagnostic). Two
+  new `pub const`s let embedders size their runtime: `MAX_RECURSION_DEPTH` and
+  `RECOMMENDED_STACK_SIZE` (16 MiB). The guard only fires *before* an overflow
+  on a thread that meets that floor — the reference REPL now sizes its tokio
+  worker threads (`thread_stack_size`) and its `block_on` driver thread to it;
+  embedders should too (see `docs/EMBEDDING.md`).
 
 ### Fixed
+- **Deep recursion no longer crashes the process** (GH #46). `f() { f; }; f`,
+  mutual recursion, and deeply nested `$(...)` aborted with a bare stack
+  overflow; they now hit the depth guard above and fail loudly.
 - **`$(...)` in a redirect target now works on a bare `Kernel::execute`** (GH
   #90). Command substitution in a redirect target or heredoc body (`echo x >
   $(gen)`, `cat < $(gen)`, `cmd > $(gen)` in a pipeline stage) only ran when the
--- a/crates/kaish-help/content/en/limits.md
+++ b/crates/kaish-help/content/en/limits.md
@@ -48,6 +48,7 @@ plain command is wanted — `test -f x && echo yes`, `if test "$a" = "$b"; then`
 - **Pipeline stages run concurrently** with isolated scopes (like bash subshells). Variable assignments in one stage aren't visible in others. Last stage syncs back to parent.
 - **Scatter results in completion order**, not input order.
 - **Command substitution runs in redirect targets and here-doc bodies** — `cmd > $(gen-path)`, `cat < $(find-cfg)`, and `$(...)` inside a here-doc body all work. The target is a single word, so quote it when it mixes text with an expansion: `> "/tmp/$(id -u).log"`, not `> /tmp/$(id -u).log`.
+- **Recursion is depth-capped at 32** — nested `$(...)`, recursive shell functions, and `.kai` scripts sourcing each other are bounded so a runaway (or a missing base case) returns a loud `maximum recursion depth exceeded` error instead of overflowing the stack. Real recursion nests far shallower; this only stops runaways.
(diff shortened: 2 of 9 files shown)


tobert:

**One real gap found and fixed; everything else holds.**

### Gap: `source`/`.` was unguarded (fixed in 1cc9368)
`source`/`.` is a **fourth** dynamic re-entry point — intercepted as a special form in `execute_command_depth` *before* the three guarded paths, and `execute_source` loops `execute_stmt_flow` inline with no `enter_recursion`. So a self-sourcing `.kai` (`source self.kai`) still SIGSEGV'd (verified: exit 134 even on the 16 MiB driver thread). Added the guard at the top of `execute_source`; now loud (exit 1), with a regression test driving a self-sourcing script through the in-memory VFS. kaibo confirmed no *other* re-entry escapes the guard: `timeout` re-dispatches through `dispatch_command` (hits the guarded sites), `exec` replaces the process, and no `eval_*` re-enters except via the already-guarded `execute_block_capturing`.

### Confirmed sound
- **RAII balance** — guard built before the ceiling check, so the error path drops-and-decrements; cancellation/unwind drop it too. Balanced on every path.
- **Fork reset correct** — a fork gets a fresh counter *and* a fresh OS stack; inheriting the parent's depth would waste a fork's full 16 MiB.
- **`try_execute_script` transient increment** — two `Relaxed` atomics per external command, harmless; counter is per-kernel and serialized by `execute_lock`.
- **REPL on a spawned `std::thread`** — safe: tokio's signal thread is internal, `init_terminal`/`tcsetpgrp` and rustyline raw-mode act on the process/fd not the thread, and the original main thread only inits tracing then `join()`s.
- **32 / 16 MiB internally consistent** — 32 levels = ~12.2 MB debug / ~2.6 MB release, leaving ~3.8 MB (≈10 levels) debug headroom for the non-recursion frames.

4379 suite + clippy `--all-targets` clean after the fix.

PR #122, opened 2026-07-06
fix(glob): pattern passthrough — glob **/*.rs no longer eats its own pattern

tobert:
## Problem

Live repro at the REPL: `glob **/*.rs` at the kaish repo root took a long time, then printed exactly one file — `crates/kaish-client/src/embedded.rs`.

Contributing factors, in order:

1. **The argv binder pre-expands bare glob patterns.** An unquoted `**/*.rs` is a `GlobPattern` token, and `build_args_async` expands it into matching paths before the tool runs. Correct shell semantics for `cat *.rs` — fatal for a tool whose *input is the pattern*.
2. **The builtin silently dropped everything past positional 0.** Its clap struct declares `pattern: Vec<String>` ("Glob pattern(s) to expand"), but execute read only the first — so the alphabetically-first pre-expanded path became the "pattern", an all-literal glob that matches exactly itself.
3. **The walk was paid twice** (bind-time expansion, then the builtin's own walk), unfiltered through `target/` under the REPL's `IgnoreConfig::none()` — that was the "long time".

The builtin's own schema examples teach the unquoted spelling, so agents following the help walk straight into this.

## Decision

A first-class seam, not a special case: `ToolSchema` grows **`glob_passthrough`** (sibling of `raw_argv`/`owns_output`) telling the binder to hand bare patterns through as written — the existing eval fallback already binds `Expr::GlobPattern` to its literal text (it's the path `set +o glob` always used). `glob` opts in, and now consumes **all** positionals as patterns: deduped union in pattern order, strict per-pattern no-match errors that name the pattern that missed.

Builtins owning pattern semantics is the established shape — `cat`/`head`/`tail`/`ls`/… already re-expand string args via `ctx.expand_paths` — the binder just stops pre-chewing glob's input. Embedder tools with pattern-shaped inputs get the same opt-in (additive field on the non_exhaustive struct).

## Verified

- 9 new kernel-routed tests (`glob_pattern_passthrough_tests.rs`): unquoted/quoted parity, multi-pattern union + dedupe, per-pattern no-match naming, noglob mode, and regression guards that other builtins keep shell expansion + strict no-match.
- End-to-end via the real REPL (`-c` and the interactive loop): the original repro now returns all 295 `.rs` files at the repo root in 0.27s.
- `cargo test --all` and `cargo clippy --all --all-targets` clean; `regen_syntax` run (drift test green).

## Follow-up (out of scope)

The REPL's `IgnoreConfig::none()` still walks `target/` unfiltered on any repo-root recursive glob — correct but slow on cold cache. Worth its own design discussion (REPL default vs `agent()` config).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,22 @@ breaking entries are marked **BREAKING**.
 
 ## [Unreleased]
 
+### Fixed
+- **Unquoted `glob **/*.rs` now works: the pattern reaches the builtin as
+  written.** Previously the kernel's argv glob expansion pre-expanded the bare
+  pattern into matching paths, so `glob` bound the first path as its "pattern",
+  silently ignored the rest, and printed exactly one file — after walking the
+  tree twice. The builtin's own examples (and agents following them) spell the
+  pattern unquoted. Quoted patterns behave as before.
+
 ### Added
+- **`ToolSchema::glob_passthrough` (+ `with_glob_passthrough()`)** — a tool
+  whose input *is* a glob pattern (like `glob`) can now tell the argv binder to
+  pass bare patterns through as literal text instead of expanding them.
+  Embedder tools with pattern-shaped inputs can opt in the same way.
+- **`glob` accepts multiple patterns** (`glob **/*.rs **/*.toml`): matches are
+  the deduped union in pattern order; any pattern with zero matches fails the
+  whole command (exit 1) naming the pattern that missed.
 - **A backgrounded confirmation latch is now surfaced and fulfillable** (GH
   #96). A destructive op gated under `set -o latch` and run in the background
   (`rm x &`) stored its `LatchRequest` but no consumer exposed it, so the nonce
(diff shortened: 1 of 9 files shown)


tobert:
**kaibo review (cast: deepseek)** — verdict: correct, no blocking issues.

Verified by the reviewer: dedupe-by-reported-name is collision-safe across anchored/relative mixes (both roots resolve identically in every current backend); all 4 `expand_glob` call sites accounted for (for-loop and external-command paths correctly have no schema, so no passthrough applies); `set +o glob` behavior unchanged and tested; quoted patterns unchanged (`Expr::String` never hits the GlobPattern arm); serde attrs match the `raw_argv`/`owns_output` wire conventions (flat tools serialize identically).

**Pre-existing nits surfaced (deferred, not from this change):**
1. `glob.rs` `no_ignore` read: `has_flag("no-ignore") || has_flag("no-ignore")` — same literal twice; second was presumably meant to be the snake-case alias.
2. `--include`/`--exclude` are declared repeatable in help but bound as `Option<String>` — only the last wins; with multi-pattern support, repeatable filters are the natural expectation.
3. `--depth` read from raw args instead of the parsed clap struct (the grep/rg context-flag bug class from CLAUDE.md).

These plus the REPL `IgnoreConfig::none()` target/-walk follow-up from the PR body are candidates for `gh issue create` if not folded into a cleanup pass.

PR #123, opened 2026-07-07
fix(#96): jobs --cleanup and kill %N no longer destroy a latch silently

tobert:
## What

A pre-release fishing sweep (gemini-pro batch + deepseek consult over the whole scheduler/jobs surface, findings verified against the binary) caught two holes in the backgrounded-latch guarantee #96 just shipped. The stored `LatchRequest` is the **only** handle to a gated background operation, and both job-dropping consumers destroyed it without a word:

- `jobs --cleanup` reaped a latched job like any completed one (`is_done()` counts `Latched` as done) — verified: "Cleaned up 1 completed job(s)", latch gone, operation permanently unconfirmable.
- `kill %N` ran `cancel()`+`remove()` unconditionally — verified: no warning, no trace.

## Decisions

- **`jobs --cleanup` keeps latched jobs** and reports `Kept N latched job(s)` so the survivor is loud, not a surprise in the next `jobs` listing.
- **`kill %N` refuses a latched job** (exit 1, points at `/v/jobs/N/latch` and the escape hatch) rather than being repurposed as the discard path — fat-fingering `kill` at a gate you meant to confirm was exactly the failure scenario found.
- **`kill --discard %N` is the escape hatch**: explicit, loud about what it abandoned, and the gated operation never runs.
- **Rider:** `Job::try_poll`'s "shouldn't happen" `Pending` branch now puts the taken `JoinHandle` back instead of dropping it — if the runtime ever returns `Pending` after `is_finished()`, the old code stranded the job as Running forever with its result silently lost.

New guard tests extend the #96 section of `latch_trash_tests.rs` (cleanup keeps + still confirmable end-to-end; kill refuses + still confirmable; discard is loud and the gated `rm` never runs). Gates: `cargo test --all` + `cargo clippy --all --all-targets` clean.

Related surfacing residuals (wait --json shape, jobs --json latch column, scatter rows, mid-pipeline latch policy) are being filed as issues rather than stretched into this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,14 @@ breaking entries are marked **BREAKING**.
   variants/fields won't break you.
 
 ### Fixed
+- **A backgrounded confirmation latch can no longer be destroyed silently**
+  (GH #96 follow-up). `jobs --cleanup` used to reap a latched job like any
+  completed one, and `kill %N` removed it outright — both silently dropped the
+  stored `LatchRequest`, leaving the gated operation permanently
+  unconfirmable. `jobs --cleanup` now keeps latched jobs and says so; `kill
+  %N` on a latched job refuses with a pointer to the nonce; the new `kill
+  --discard %N` abandons the gate explicitly and loudly (the gated operation
+  never runs).
 - **README install instructions pointed at a crate that doesn't exist.**
   `cargo install kaish` fails — there is no `kaish` package on crates.io; the
   binary named `kaish` ships in the `kaish-repl` crate. The README now says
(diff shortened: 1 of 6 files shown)


tobert:

Scrutinized: the is_latched→kill_job TOCTOU (benign — the guard protects an already-visible confirmation from accidental destruction; killing a still-running job is stated intent), whether refusing kill breaks teardown flows (no — the refusal names `--discard`), every other job-dropping path (fg/bg structurally can't reach a latched job: it has no pid/pgid and isn't stopped; `Kernel::shutdown` waits but doesn't remove; `JobManager::remove` is the one unguarded public seam), `--discard` × signal/pgid/hermetic interactions (the guard short-circuits before all of them), and the try_poll fix (correct, cheap, defensive).

Follow-ups it suggested are folded into the head commit: TOCTOU rationale comment, `remove()` doc warning, `--discard` conflicts-with `--signal`, and edge-case tests (nonexistent job, non-latched job, flag conflict).

Issue #135, opened 2026-07-07
flaky test: external_command_tests::non_interactive_stdin_is_dev_null fails under full-suite load


tobert:
Pre-existing flake, first noted in the #48 session and hit again 2026-07-06 during an unrelated PR's `cargo test --all` run: `non_interactive_stdin_is_dev_null` failed with `stdin should be /dev/null: socket:[...]` under peak parallel load, then passed in isolation (and historically passes full-binary and 4× reruns). The `#[ignore]`'d sibling test notes the same stdin-harness sensitivity.

Worth a flake-hardening pass: either serialize the stdin-sensitive externals behind a mutex, or make the assertion tolerant of the harness's socket-stdin when the test runner itself owns fd 0.

tobert:
Spent a genuine effort trying to reproduce this under load; could not, despite pushing well past the conditions under which it was originally seen. Leaving this open with the evidence rather than shipping a speculative fix (mutex-serialize vs. loosened assertion) with nothing to verify it against.

**What was tried** (branch `fix/flaky-stdin-test-135`, no code changes made):

1. Baseline: `cargo test -p kaish-kernel --test external_command_tests --features subprocess -- --test-threads=64` — 1 run, clean (target test `ok`).
2. `hammer.sh`: 15x repeats of the same isolated-binary invocation via `cargo test` — 15/15 clean.
3. Direct cross-process hammering (bypassing cargo, invoking the compiled `external_command_tests` binary directly, many concurrent OS processes at once, repeated in rounds):
   - 8 rounds x 8 concurrent copies x `--test-threads=32` = 64 copy-runs
   - 40 rounds x 24 concurrent copies x `--test-threads=96` = 960 copy-runs
   - 60 rounds x 32 concurrent copies x `--test-threads=128` = 1920 copy-runs
   - 50 rounds x 48 concurrent copies x `--test-threads=16` = 2400 copy-runs
   - 30 rounds x 64 concurrent copies x `--test-threads=32` = 1920 copy-runs
   - **7264/7264 copy-runs clean** — zero failures of the target test at any concurrency/thread-count combination tried.
4. Full-suite `cargo test --all --features subprocess -- --test-threads=<N>` (closest match to how the flake was originally observed):
   - One run at `--test-threads=128`: fully clean (grepped the whole log for `FAILED|panicked|should be /dev/null` — no hits).
[...]

PR #136, opened 2026-07-07
fix(walk-filters): --include works, repeats accumulate, bad values go loud

tobert:
## What

A fishing sweep for siblings of #122's bug classes (flags read off the raw `ToolArgs` map instead of the parsed clap struct; `Option<String>` numerics with `parse().ok()` fallbacks) found the walk-filter family broken three ways. Every finding was verified against the binary before fixing:

1. **`glob --include` never filtered at all.** `IncludeExclude::should_exclude` only acted on explicit Exclude rules — the "strict mode" its own doc comment described was never implemented. (`grep --include` only appeared to work because it separately baked the single include into its walk pattern.)
2. **Repeating `--include`/`--exclude` silently kept only the last value** in glob and grep — glob's help even claimed "can be repeated", and `grep -r TODO . --include='*.rs' --include='*.toml'` silently searched only toml (a false negative).
3. **Malformed numeric values fell back silently**: `glob --depth=abc` / `tree -L abc` / `find -maxdepth xyz` walked unlimited at exit 0; `spawn timeout=abc` silently **disabled** the timeout; `split --limit` and `head -c` wrapped negatives through `usize`.

## Decisions

- The filter now owns include semantics via a single entry-aware `excludes_entry()`: a file must match an include when includes exist (relative path or basename), **directories are never pruned by include-miss** (traversal still reaches included files below), and excludes still prune subtrees. The walker's two-call check collapsed into one; grep's `**/{inc}` pattern hack is deleted.
- Repeatables bind `Vec<String>` and read via `read_repeatable_strings` — the `--ftype` pattern.
- Numerics moved to clap-validated `Option<i64>` (or loud hand parses where the argv is domain-shaped: find, spawn); negatives are refused, not wrapped. `glob --type=<unknown>` (silently "files only") is now a usage error too.
- Remaining raw-map reads of clap-declared flags in these builtins moved to the parsed struct per the CLAUDE.md convention — which also deletes the `no-ignore || no-ignore` copy-paste dupes in glob and tree.

New kernel-routed suite `walk_filter_flags_tests.rs` (11 tests) pins all of it, including the traversal-vs-include-miss distinction. Gates: full tests green (one unrelated known flake, #135) and clippy `--all --all-targets` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,26 @@ breaking entries are marked **BREAKING**.
 ## [Unreleased]
 
 ### Fixed
+- **`glob --include` now actually filters.** It was a complete no-op: the
+  walker consulted only exclude rules, so `glob '*' --include='*.rs'` listed
+  everything. Include semantics are now rg-like: when include patterns exist a
+  file must match one of them (by relative path or basename); directories are
+  still traversed so included files below them are reached; exclude patterns
+  still prune whole subtrees. (`grep --include` half-worked through a separate
+  walk-pattern hack, now removed in favor of the same filter.)
+- **Repeated `--include`/`--exclude` accumulate in `glob` and `grep`.** Both
+  were bound single-valued, so repeating the flag silently kept only the LAST
+  pattern — `glob` while its help said "can be repeated", and `grep -r TODO .
+  --include='*.rs' --include='*.toml'` silently searched only the toml files
+  (a false negative). Repeats now accumulate like `--ftype` always has.
+- **A malformed numeric flag value errors instead of silently meaning
+  "unlimited"/"disabled".** `glob --depth=abc`, `tree -L abc`, and
+  `find -maxdepth xyz` all exited 0 and walked without a depth limit;
+  `spawn timeout=abc` silently DISABLED the timeout (unbounded child);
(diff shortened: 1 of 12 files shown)


tobert:

Scrutinized: existing `IncludeExclude` callers (only glob and grep construct one; every other walker consumer gets the empty default — no behavior change), the relative-then-basename ordering in `excludes_entry` (sound; the only surprise case comes from `CompiledRule`'s pre-existing filename fallback, unchanged by this PR), the deleted grep `**/{inc}` pattern hack (dotfile reach and I/O unchanged — both paths were `**`-prefixed globs; multi-include is strictly new capability), redundant `glob '**/*.rs' --include='*.rs'` combos (both gates agree; contradictory combos hit the strict no-match error), spawn Float timeouts (now loud instead of silently disabling — strict improvement), and the negative-value refusals against GNU (`head -n -N` still works; `head -c -N` is refused with a clear message, acceptable under the 80% rule).

Follow-ups folded into the head commit: the review's exit-code nit (head negative `-c` now exits 2 like the rest of the family), plus a fish the spawn test caught on its first feature-gated run — **spawn's own help examples taught a form that exits 127** (`command="cargo"` word-assign never binds for spawn); examples now show the working `--command` flag form, with a CHANGELOG bullet.

PR #138, opened 2026-07-07
feat(#134): the reference REPL is ignore-aware by default

### LINKED ISSUE #134: REPL defaults to IgnoreConfig::none() — interactive walks traverse target/, .git/, node_modules (design call)
From the 2026-07-06 fishing sweep. `KernelConfig::repl()` (and most constructors) default to `IgnoreConfig::none()` — literally zero filtering (`kernel.rs` ~336+): no .gitignore, no default ignores. So interactive `glob '**/*.rs'`, `tree`, and `grep -r` walk `target/`, `.git/`, `node_modules/` in full. This contributed to the "glob walked forever" repro that led to #122 — the argv-expansion bug made it visible, but the unfiltered walk made it slow.

Design options (Amy's call):
1. REPL uses `IgnoreConfig::agent()` (Enforced scope + .gitignore + defaults) — predictable, but Enforced means no per-call opt-out until `--no-ignore` lands across builtins.
2. A new `interactive()` preset: .gitignore + defaults with **Suggest** scope, so `--no-ignore` can override.
3. Keep `none()` and document that interactive users should `kaish-ignore` themselves.

Option 2 seems most honest for a human at a prompt: sensible defaults, recoverable per call. Related: the standing `--no-ignore`-under-Enforced design question in docs/issues.md ("--no-ignore for the search builtins").

tobert:
## What

Amy's design call on #134: interactive kaish should default to ignore-aware walking. A new `IgnoreConfig::interactive()` preset — `agent()`'s filters (`.gitignore` + default ignore list + auto-gitignore) at **Advisory** scope — becomes `KernelConfig::repl()`'s default. glob/tree/grep/ls filter; `find` keeps its POSIX everything-visible tradition; recovery is `--no-ignore` per call or `kaish-ignore clear` per session. Embedded presets (`transient`/`named`/`isolated`) keep unfiltered `none()` — the choice belongs to the frontend.

## The bug found on the way

Pinning the opt-out path exposed a pre-existing hole: **`kaish-ignore` mutations never persisted past their own statement.** `execute_command`'s post-tool sync copies back cwd/aliases/output-limit — with a comment documenting exactly this missing-field class for output-limit — but not `ignore_config`. So `add`/`clear`/`defaults`/`scope` all silently died at the statement boundary, and the help's documented rc-file recipe (`kaish-ignore add .gitignore`) had been a no-op. One line in the sync fixes it; it's another tally for the "extract the ctx-sync helper" backlog entry (third manual sync site, second missing field).

## Verification

Five kernel-routed tests (`repl_ignore_default_tests.rs`): filtering via glob and grep, `--no-ignore` recovery, `find` unrestricted under Advisory, and `kaish-ignore clear` persisting across statements. E2E on this repo: `glob '**/*.rs'` = 294 filtered vs 298 with `--no-ignore`; `find target` still lists build artifacts. `help ignore` content updated (modes table + opt-out section). Gates: full tests green (only the known flake #135), clippy clean.

Closes #134.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,21 @@ breaking entries are marked **BREAKING**.
 
 ## [Unreleased]
 
+### Changed
+- **The reference REPL is ignore-aware by default** (GH #134). Interactive,
+  `-c`, and script modes now load `.gitignore` and the default ignore list
+  (`.git`, `target`, `node_modules`, …) at Advisory scope: glob/tree/grep/ls
+  filter, `find` stays POSIX-unrestricted, and the unfiltered view is one
+  `--no-ignore` (per call) or `kaish-ignore clear` (per session) away. A new
+  `IgnoreConfig::interactive()` preset carries this; bare embedded kernels
+  (`transient`/`named`/`isolated`) keep the unfiltered default.
+
 ### Fixed
+- **`kaish-ignore` changes now persist past their own statement.** Every
+  runtime ignore mutation (`add`/`clear`/`defaults`/`scope`) was silently
+  dropped at the end of the statement that made it — the per-command context
+  sync copied back cwd/aliases/output-limit but not the ignore config — so
+  the documented `kaish-ignore add .gitignore` rc-file recipe did nothing.
 - **`glob --include` now actually filters.** It was a complete no-op: the
   walker consulted only exclude rules, so `glob '*' --include='*.rs'` listed
   everything. Include semantics are now rg-like: when include patterns exist a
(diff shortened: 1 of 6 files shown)


tobert:

Traced every `KernelConfig::repl()` consumer (REPL/-c/script are the target; the ~15 test files using `common::kernel_at` create no ignorable fixtures; kaish-wasi uses `isolated()`, kaish-client `transient()` — unaffected), every sync path for the one-line `ignore_config` copy-back (foreground pipeline, tool dispatch, both dispatch_command directions all correct; background forks and scatter workers stay properly isolated — `(kaish-ignore clear) &` does not bleed), the struct-update inheritance (correct; inheritance note added to `agent()` per the review), and startup cost (filter builds lazily at walk time — `kaish -c 'echo hello'` pays nothing).

One pre-existing gap it found, filed as #139: `$(...)` substitution saves/restores only scope/cwd, so `$(kaish-ignore clear)` (and `$(kaish-output-limit …)`, `$(alias …)`) mutates the session — the inverse of the bug this PR fixes, same manual-sync class, more ammunition for the ctx-sync-helper extraction.

PR #169, opened 2026-07-13
ci: add GitHub Actions workflow encoding the CLAUDE.md gates

tobert:
kaish has had no CI since the repo began — every gate in CLAUDE.md (tests, clippy,
the capability-feature sandbox build, the WASI target) ran only on contributor
machines. This PR adds the repo's first GitHub Actions workflow, modeled on
kaibo's `ci.yml` but encoding kaish's own gates. kaish releases to crates.io
rather than shipping binaries, and publishing stays deliberately manual (the
`/release` runbook), so there is no release workflow — `ci.yml` is the whole story.

Three parallel jobs, all with pinned-by-SHA actions (a tag can be moved, a digest
can't):

- **test + clippy** — `cargo test --all --locked`, then
  `cargo clippy --all --all-targets --locked -- -D warnings`, then a tripwire that
  fails if any insta `*.snap.new` pending snapshot was committed (under CI insta
  fails-without-writing, so any `.snap.new` found was committed from a local run).
- **no-default-features** — `cargo test -p kaish-kernel --no-default-features
  --locked`, the gate that catches OS-touching code landing outside its capability
  axis.
- **wasi build** — `cargo build -p kaish-wasi --target wasm32-wasip1 --locked`, so
  embedders on that target don't discover breakage at release time.

Decisions along the way:

- `--locked` everywhere: `Cargo.lock` is tracked, and CI drifting to newer deps
  than contributors test against is a silent-fallback class of surprise.
- No `cargo insta test --check` job: `cargo test --all` already fails on snapshot
  mismatch under CI, and the `.snap.new` tripwire covers the committed-pending
  case — same coverage without installing cargo-insta or running the suite twice.
- kaibo's TLS-invariant guard was dropped: it exists to keep kaibo's static musl
  builds honest, and kaish has no TLS or binary-release surface.
- A kaibo (deepseek) review of the workflow found one real gap, fixed here:
  `.gitignore` had no `*.snap.new` entry, so the very mistake the CI tripwire
  catches could be staged by `git add .` locally. Belt added; CI stays suspenders.

All three legs were verified locally before encoding. This PR's own checks are the
live validation that the workflow runs on `pull_request`; a follow-up doc-sweep PR
(README badge + CI mentions in CLAUDE.md/README) validates the on-push-to-main leg
and the badge.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,101 @@
+# The correctness gates from CLAUDE.md, run on every pull request and every push to
+# main: the full workspace test suite, clippy with warnings denied (test targets
+# included), the capability-feature sandbox compile, and the WASI target build.
+# kaish releases to crates.io (no binary artifacts), and publishing is deliberately
+# manual — the /release runbook — so this is the only workflow.
+#
+# Actions are pinned by commit SHA, not tag — a tag can be moved, a digest can't
+# (supply-chain). Each pin carries a `# vX.Y.Z` comment naming the release it points
+# at, for humans and Dependabot.
+name: ci
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test:
+    name: test + clippy
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@fa04a1451ff1842e2626ccb99004d0195b455a88 # stable branch, pinned 2026-07-05
+        with:
+          toolchain: stable
+
+      - name: Cache cargo
+        uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
+
+      - name: Test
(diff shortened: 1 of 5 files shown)


tobert:
## First live run: two failures, both real

The first CI run on this PR failed two of three jobs — and both failures were
pre-existing repo bugs the workflow correctly exposed, not workflow bugs:

**`test + clippy`: `non_interactive_stdin_is_dev_null` flake.** The test's comment
says "Use /bin/readlink to bypass the builtin," but the code ran bare `readlink` —
which resolves to kaish's *builtin* readlink. The builtin resolves
`/proc/self/fd/0` of the **test process**, not of a spawned child, so the test
only ever passed in environments where the test runner itself had
stdin=/dev/null. GitHub's runner gives the step a pipe → `pipe:[17586]` →
failure. Reproduced locally with `echo poke | cargo test …` (red), fixed by
making the code match its comment (`/bin/readlink`), re-ran with piped stdin
(green). The ignored `interactive_stdin_is_not_dev_null` got the same fix for
intent-consistency.

**`no-default-features`: never actually compiled.** `KernelConfig::repl()` is
`#[cfg(feature = "localfs")]` and ~25 integration-test files call it —
`cargo test -p kaish-kernel --no-default-features` has never built. My earlier
"local verification" of this leg piped cargo through `tail`, so the exit code I
checked was tail's, not cargo's (mea culpa; rerun properly it exits 101). The
documented invariant is "sandbox-mode **compiles**", so the leg is now
`cargo check --no-default-features` — and upgrading it to `cargo test` by
feature-gating the test crates is filed as #170.

PR #238, opened 2026-07-18
feat(types): persist OutputData.rich_json through serde

tobert:
## What

Flip `OutputData.rich_json` from `#[serde(skip)]` to
`#[serde(default, skip_serializing_if = "Option::is_none")]` so it persists
through self-describing serde formats (JSON, CBOR).

## Why

`rich_json` is the `--json` render override (set by `grep --json`, available to
any builtin). It was skipped as a "transient render hint." The parenthetical
reason was postcard/bincode safety — `serde_json::Value` can't deserialize from
non-self-describing formats (no `deserialize_any`), so skipping kept
`OutputData` safe for them.

**But there's no such path.** No `postcard::`/`bincode::` call touches
`OutputData` or `ExecResult` anywhere in kaish, and the embedder that persists
them (kaijutsu) is CBOR — self-describing, where `serde_json::Value` round-trips
fine. The skip cost embedders the ability to persist a builtin's rich `--json`
output. Concretely: kaijutsu wires a `kj` command's structured payload onto
`rich_json` to reach its MCP surface, and serde silently dropped it before the
block was stored — so MCP agents saw `null`/empty where the structured data
should be. Discovered while wiring that surface.

## The change

- serializes when `Some` -> persists through JSON/CBOR
- omitted when `None` (the common case) via `skip_serializing_if` -> wire shape
  unchanged, and a `None` stays safe even for non-self-describing formats
- `default` handles absence on decode

**Caveat** (documented on the field + CHANGELOG): don't add a postcard/bincode
serializer for `OutputData` while `rich_json` may be `Some`.

## Tests

- `rich_json_round_trips_through_serde` — serialize -> deserialize keeps it
- `rich_json_none_stays_off_the_wire` — `skip_serializing_if` omits `None`

`cargo test --all --locked` green, `cargo clippy --all-targets -D warnings` clean.

A kaibo review will follow as a comment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,18 @@ breaking entries are marked **BREAKING**.
 
 ## [0.13.0] - 2026-07-18
 
+### Added
+- **`OutputData.rich_json` now persists through serde** (self-describing
+  formats: JSON, CBOR). The `--json` render override — set by `grep --json`,
+  available to any builtin — was `#[serde(skip)]` as a transient render hint.
+  It now serializes when `Some` (omitted when `None` via
+  `skip_serializing_if`), so an embedder can carry a builtin's rich structured
+  output onto a stored record and read it back — e.g. kaijutsu persisting a
+  `kj` command's structured output onto a CRDT block for its MCP surface. The
+  `None` (common) case is unchanged on the wire. Caveat: a `Some` value can't
+  round-trip through non-self-describing formats (postcard/bincode lack
+  `deserialize_any`) — kaish uses neither, and kaijutsu is CBOR.
+
 ### Fixed
 - **Arg-binding polish** (GH #189): four small gaps in the shared arg binder
   (`kernel::bind_tool_args`), verified against current code post-#188/#231:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,9 +20,11 @@ breaking entries are marked **BREAKING**.
   `skip_serializing_if`), so an embedder can carry a builtin's rich structured
   output onto a stored record and read it back — e.g. kaijutsu persisting a
   `kj` command's structured output onto a CRDT block for its MCP surface. The
-  `None` (common) case is unchanged on the wire. Caveat: a `Some` value can't
-  round-trip through non-self-describing formats (postcard/bincode lack
-  `deserialize_any`) — kaish uses neither, and kaijutsu is CBOR.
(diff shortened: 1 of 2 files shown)


tobert:
## kaibo review (gemini-pro + deepseek)

Two-model deliberate pass. (The deepseek *agent* couldn't crawl the repo — kaibo's boundary is the embedder tree — so both ran toolless on the change + rationale; the empirical claims were verified directly: `grep` finds zero `postcard::`/`bincode::` serializations of `OutputData`/`ExecResult` in kaish, and `cargo test --all --locked` passes with no snapshot breakage.)

### Actioned (commit 25999ab)
- **Dropped `#[cfg_attr(feature = "schema", schemars(skip))]`** — both models flagged it: now that the field serializes, the generated schema shouldn't hide it. schemars emits a permissive any-value schema for `serde_json::Value`; no schema snapshot exists, and `--features schema` still builds clean.
- **Sharpened the caveat** (field doc + CHANGELOG) — gemini correctly noted the failure is on *decoding* a `Some` value from a non-self-describing format (`serde_json::Value` → `deserialize_any`), not on "adding a serializer."

### Considered and not taken
- **deepseek: "drop `Eq`"** (its headline "change-needed"). Premise was that `serde_json::Value` isn't `Eq` in 1.0 — **false**: `serde_json-1.0.150/src/number.rs:50` has `impl Eq for N {}`, JSON can't represent `NaN` (`Number::from_f64(NaN) == None`), and the derive compiles + 103 tests pass. Dropping `Eq` would be a gratuitous breaking change to a public type. gemini independently confirmed `Eq` is valid here.
- **gemini: `Option<Box<serde_json::value::RawValue>>`** (its "better yet"). It would sidestep the postcard/bincode caveat, but `RawValue` implements neither `Clone` nor `PartialEq`/`Eq`, so it would **strip those derives off `OutputData`** (a heavily-used public type) — a real breaking change for a purely theoretical benefit (there is no live non-self-describing path). Keeping `Value` is the right trade.
- **gemini: `preserve_order` risk** — kaish does enable `serde_json/preserve_order` (`IndexMap`), but `IndexMap`'s `PartialEq` is order-independent and round-trip preserves order, so `rich_json` equality stays deterministic. No action.

Both agreed the serde attributes are correct and the `None`-case wire shape is byte-identical to the old `#[serde(skip)]`.

*Reviewers: gemini-pro-latest, deepseek-v4-pro (via kaibo).*

